PR #7, opened 2015-06-25
update engine config to allow users to add elements to under key extensions

ABaldwinHunter:
@codeclimate/review This PR adds the ability for users to add extensions under an `extensions` key in their eslint engine in .codeclimate.yml file. It keeps `.js` as a default extension if none is provided.

### DIFF
--- a/bin/eslint
+++ b/bin/eslint
@@ -37,6 +37,9 @@ if (process.env.ENGINE_CONFIG) {
   if (engineConfig['exclude_paths']) {
     ignores = engineConfig['exclude_paths'];
   }
+  if (engineConfig['extensions']) {
+    options['extensions'] = engineConfig['extensions'];
+  }
 }
 var cli = new CLIEngine(options);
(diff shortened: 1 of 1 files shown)


bin/eslint
```diff
@@ -37,6 +37,11 @@ if (process.env.ENGINE_CONFIG) {
   if (engineConfig['exclude_paths']) {
     ignores = engineConfig['exclude_paths'];
   }
+  if (engineConfig['extensions']) {
+    options['extensions'] = engineConfig['extensions'];
+  } else {
+    options['extensions'] = ['.js'];
```
GordonDiggs: You can leave this above, where it was. With the change as it is, if there is no engine config, we will not set an extensions option

ABaldwinHunter:
@GordonDiggs thanks for review. updated to keep `extensions: [.js]` above

GordonDiggs:
:+1: Looks good!

Issue #10, opened 2015-06-26
Doesn't load project plugins


tmoitie:
My eslintrc file is trying to use the react plugin which is not loaded by the analyzer engine

`error: (CC::Analyzer::Engine::EngineFailure) engine eslint failed with status 1 and stderr "module.js:338\n    throw err;\n          ^\nError: Cannot find module 'eslint-plugin-react'\n    at Function.Module._resolveFilename (module.js:336:15)\n    at Function.Module._load (module.js:278:25)\n    at Module.require (module.js:365:17)\n    at require (module.js:384:17)\n    at /usr/src/app/node_modules/eslint/lib/cli-engine.js:107:26\n    at Array.forEach (native)\n    at loadPlugins (/usr/src/app/node_modules/eslint/lib/cli-engine.js:98:21)\n    at processFile (/usr/src/app/node_modules/eslint/lib/cli-engine.js:179:5)\n    at /usr/src/app/node_modules/eslint/lib/cli-engine.js:332:26\n    at /usr/src/app/node_modules/eslint/lib/util/traverse.js:61:17\n"
`

GordonDiggs:
Thanks! I'm putting this in our roadmap and will fix soon

tmoitie:
Hullo. Has this PR made it into release? If so how do I update?

Thanks!

GordonDiggs:
@tmoitie We had to revert it because we actually can't support plugins just yet. I'm going to leave this issue open for now and will update it when we push support for them!

jdelStrother:
In the meantime, any suggestions on continuing to use `eslint-plugin-react` locally, but disabling it when running under codeclimate?   Can I point codeclimate's engine to a different .eslintrc or something?

jakemhiller:
Yeah, Having the same issue. Along with the react plugin we also use the [babel-eslint](https://github.com/babel/babel-eslint) parser, which was mentioned in another issue. It would be nice to at least have a work-around like what @jdelStrother mentioned.

mattdaw:
:+1: My project is blocked by this as well.

abritinthebay:
Ditto. At the very least supporting the react plugin will allow support of all the react code coming down the pipe.

While ESlint killed support of React semantics in JSX (which makes sense, tho I wish they'd just have a JSX plugin at that point) this is basically required for any JSX work that uses components (ie - most of it)

GordonDiggs:
[...]

PR #12, opened 2015-06-30
Allow plugins to be loaded from /code/node_modules

GordonDiggs:
/cc @codeclimate/review 

This fixes the recurring bugsnag as well as #10

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,4 +10,6 @@ USER app
 
 COPY . /usr/src/app
 
+ENV NODE_PATH /code/node_modules
+
 CMD ["/usr/src/app/bin/eslint.js"]
--- a/package.json
+++ b/package.json
@@ -8,8 +8,7 @@
     "url" : "http://github.com/codeclimate/codeclimate-eslint.git"
   },
   "dependencies": {
-    "eslint": "0.22.1",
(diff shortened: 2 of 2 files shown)


pbrisbin:
Doesn't this just fix the specific plugin? Don't we need to support the user having their own plugins in `./node_modules`?

GordonDiggs:
@pbrisbin ready for another look

Dockerfile
```diff
@@ -10,4 +10,6 @@ USER app
 
 COPY . /usr/src/app
 
+ENV NODE_PATH /code/node_modules
```
pbrisbin: Should this be something like `/code/node_modules:/usr/src/app/node_modules`?

pbrisbin: and even `:/whatever/the/system/node_modules` too?

GordonDiggs: Nope - by default it is nothing and is only used when the module can't be found elsewhere:

> If the NODE_PATH environment variable is set to a colon-delimited list of absolute paths, then node will search those paths for modules if they are not found elsewhere.

*- https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders*

pbrisbin: Cool!

pbrisbin:
LGTM

PR #13, opened 2015-06-30
Add circle.yml file to build docker images on push

GordonDiggs:
@pbrisbin can you review this? I'll also need to know the values for the various environment variables (they are hidden in the CircleCI interface)


pbrisbin:
Ah I forgot that they were hidden. You can get them from here:

https://console.developers.google.com/project/code-climate/apiui/credential?authuser=1

Or I can give them to you over some secure channel?

PR #19, opened 2015-07-17
Add remediation points

GordonDiggs:
/cc @codeclimate/review


pbrisbin:
I assume this is the base from the SPEC? If so, LGTM

GordonDiggs:
That is correct!

PR #20, opened 2015-07-20
Update eslint to our fork

GordonDiggs:
/cc @codeclimate/review 

[These ESLint commits](https://github.com/codeclimate/eslint/compare/v0.24.1...c5457c8?diff=unified&name=c5457c8) disable requiring of plugins and shared configs, until we can support them in the future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@ FROM mhart/alpine-node
 WORKDIR /usr/src/app
 COPY package.json /usr/src/app/
 
+RUN apk --update add git
+
 RUN npm install
 
 RUN adduser -u 9000 -D app
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "url" : "http://github.com/codeclimate/codeclimate-eslint.git"
   },
   "dependencies": {
-    "eslint": "0.24.1",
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -3,6 +3,9 @@ FROM mhart/alpine-node
 WORKDIR /usr/src/app
 COPY package.json /usr/src/app/
 
+RUN apk --update add git build-base && \
+    apk del build-base && rm -fr /usr/share/ri
```
pbrisbin: does this line install, then immediately remove `build-base`?

GordonDiggs: Yes we do that everywhere. [Another example](https://github.com/codeclimate/codeclimate/blob/master/Dockerfile#L11)

pbrisbin: That example needs `build-base` present for the `bundle` step in the middle. This doesn't seem to fit that.

GordonDiggs: Ahh - will try removing it

pbrisbin: The `rm /usr/share/ri` is also because of the `bundle` step in the middle. I think you just want to add `git` here and that's all.

GordonDiggs: Done!

pbrisbin:
LGTM

PR #29, opened 2015-09-09
Update to handle config.include_paths.

fhwang:


GordonDiggs:
LGTM

One thing to note: it is possible that someone will install engines without updating the CLI. Do you think it's worth it to exit early if `include_paths` were not passed?

fhwang:
Seems like we should handle in some way or another. How have we been thinking about CLI versioning? Would this require a bump to `0.4.0` in `codeclimate`, and then we could issue a warning here?

fhwang:
Obviated by #30 .

PR #30, opened 2015-09-10
Update to handle config.include_paths or config.exclude_paths.

fhwang:
For the time being, engines need to support both the old `include_paths` and the new `exclude_paths` keys in `config.json`. This PR represents that requirement.

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -57,10 +57,8 @@ function exclusionBasedFileListBuilder(excludePaths) {
 
     allFiles.forEach(function(file, i, a){
       if(excludePaths.indexOf(file.split("/code/")[1]) < 0) {
-        if(!fs.lstatSync(file).isDirectory()) {
-          var extension = "." + file.split(".").pop();
-
-          if(extensions.indexOf(extension) >= 0) {
+        if(fs.lstatSync(file).isFile()) {
+          if (isFileWithMatchingExtension(file)) {
             analysisFiles.push(file);
           }
         }
@@ -69,11 +67,6 @@ function exclusionBasedFileListBuilder(excludePaths) {
 
     return analysisFiles;
   };
-}
-
-function isFileWithMatchingExtension(file, extensions) {
-  var extension = "." + file.split(".").pop();
-  return (extensions.indexOf(extension) >= 0);
 }
 
 function inclusionBasedFileListBuilder(includePaths) {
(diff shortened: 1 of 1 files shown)


bin/eslint.js
```diff
@@ -45,33 +45,72 @@ function buildIssueJson(message, path) {
...
-          analysisFiles.push(file);
+function exclusionBasedFileListBuilder(excludePaths) {
+  // Uses glob to traverse code directory and find files to analyze,
+  // excluding files passed in with by CLI config, and including only
+  // files in the list of desired extensions.
+  //
+  // Deprecated style of file expansion, supported for users of the old CLI.
+  return function(extensions) {
+    var analysisFiles = [];
+    var allFiles = glob.sync("/code/**/**", {});
```
GordonDiggs: Isn't this the exact call that errors when something has restrictive permissions?

GordonDiggs: EDIT: nevermind this is the deprecated function

pbrisbin: I think that's why it's in the `exclusionBasedFileListBuilder`, this is the deprecated, backwards-compatible path so it would still retain the bug -- as long as `inclusionBasedFileListBuilder` doesn't.

bin/eslint.js
```diff
@@ -45,33 +45,72 @@ function buildIssueJson(message, path) {
...
+    var analysisFiles = [];
+
+    includePaths.forEach(function(fileOrDirectory, i) {
+      if ((/\/$/).test(fileOrDirectory)) {
+        // if it ends in a slash, expand and push
+        var filesInThisDirectory = glob.sync(
+          "/code/" + fileOrDirectory + "/**/**"
+        );
+        filesInThisDirectory.forEach(function(file, j){
+          if(!fs.lstatSync(file).isDirectory()) {
```
GordonDiggs: I've been bitten by this before. This will return true for symlinks, which will then probably cause an error. If there's an `isFile` check, we should use that.

fhwang: Ooh, good catch, thanks.

bin/eslint.js
```diff
@@ -45,33 +45,74 @@ function buildIssueJson(message, path) {
...
-  allFiles.forEach(function(file, i, a){
-    if(excludePaths.indexOf(file.split("/code/")[1]) < 0) {
-      if(!fs.lstatSync(file).isDirectory()) {
-        var extension = "." + file.split(".").pop();
-
-        if(extensions.indexOf(extension) >= 0) {
-          analysisFiles.push(file);
+function isFileWithMatchingExtension(file, extensions) {
+  var stats = fs.lstatSync(file);
+  var extension = "." + file.split(".").pop();
```
pbrisbin: Any considering to be made for (e.g.) `foo.tar.gz` or `index.html.slim`?

fhwang: This line returns the last fragment in the string, i.e. `.gz` or `.slim`. So I think this is what we want. Even in a case where `js` is in the middle of the string, such as `application.js.erb`, we wouldn't to consider that analyzable by ESLint.

pbrisbin:
LGTM

PR #32, opened 2015-10-06
Warn via STDERR when ESLint is ignoring a path included by `.codeclimate.yml`

dblandin:
ESLint emits a warning when a path, ignored with `.eslintignore`, is passed explicitly for analysis. The issue formatted by this engine doesn't pass validation and causes the engine to error out. This PR delegates to the CLIEngine's `#isPathIgnored` function to determine whether a file path is ignored by `.eslintignore`. If so, we'll write a warning out to STDERR instead of formatting an issue.

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -139,21 +139,26 @@ runWithTiming("engineConfig", function () {
   }
 });
 
+var cli = runWithTiming("cliInit", function() { return new CLIEngine(options); });
+
 var analysisFiles = runWithTiming("buildFileList", function() {
   return buildFileList(options.extensions);
 });
-var cli = runWithTiming("cliInit", function() { return new CLIEngine(options); }),
-    report = runWithTiming("cliRun", function() { return cli.executeOnFiles(analysisFiles); });
+
+var report = runWithTiming("cliRun", function() { return cli.executeOnFiles(analysisFiles); });
 
 runWithTiming("resultsOutput",
   function() {
     report.results.forEach(function(result) {
       var path = result.filePath.replace(/^\/code\//, "");
-
-      result.messages.forEach(function(message) {
-        var issueJson = buildIssueJson(message, path);
-        console.log(issueJson + "\u0000");
-      });
+      if (cli.isPathIgnored(path)) {
+        process.stderr.write("File `" + path + "` ignored because of your .eslintignore file." + "\u0000")
+      } else {
+        result.messages.forEach(function(message) {
+          var issueJson = buildIssueJson(message, path);
+          console.log(issueJson + "\u0000");
+        });
+      }
     });
   }
 );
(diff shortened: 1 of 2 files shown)


bin/eslint.js
```diff
@@ -139,21 +139,26 @@ runWithTiming("engineConfig", function () {
...
   function() {
     report.results.forEach(function(result) {
       var path = result.filePath.replace(/^\/code\//, "");
-
-      result.messages.forEach(function(message) {
-        var issueJson = buildIssueJson(message, path);
-        console.log(issueJson + "\u0000");
-      });
+      if (cli.isPathIgnored(path)) {
+        process.stderr.write("File `" + path + "` ignored because of your .eslintignore file." + "\u0000")
```
wfleming: We don't really have a spec for stderr (yet), but since stderr is often meant to be more human readable, this should probably get a newline instead of a null byte.

bin/eslint.js
```diff
@@ -139,21 +139,26 @@ runWithTiming("engineConfig", function () {
...
 runWithTiming("resultsOutput",
   function() {
     report.results.forEach(function(result) {
       var path = result.filePath.replace(/^\/code\//, "");
-
-      result.messages.forEach(function(message) {
-        var issueJson = buildIssueJson(message, path);
-        console.log(issueJson + "\u0000");
-      });
+      if (cli.isPathIgnored(path)) {
```
wfleming: Would it be better to do this in `buildFileList`? If the CLI will correctly answer us at that point, then it seems better to not even send it to eslint in the first place.

dblandin: If we move the check for `isPathIgnored` to `buildFileList`, I don't think we'll see this warning `result` show up, as the path will no longer be passed explicitly to `executeOnFiles`. Since we often see timeout issues with this engine already, I'm hesitant to add another loop on file paths before we hit the analysis stage. What do you think?

bin/eslint.js
dblandin: Sounds good to me! Will change this to a `"\n"` character.

bin/eslint.js
wfleming: Every log I've seen has shown that the analysis stage *is* where the timeouts occur: everything before it is plenty fast. Which is partly why I'd rather send fewer files to eslint if possible.

I don't think it needs to be an entire loop: it could go in the loops of `inclusionBasedFileListBuilder` & `exclusionBasedFileListBuilder`.

dblandin: Ahh, awesome. Okay, that sounds good to me then! I'll put that in place, further up.

dblandin:
@wfleming Ready for another :eyes:  when you have a chance!

wfleming:
LGTM

PR #33, opened 2015-10-09
Process files in batches, force garbage collection

wfleming:
This may help with our memory consumption issues with the eslint engine in production. I'm forcing GC at intervals.

I ran this locally against the [meteor/meteor](https://github.com/meteor/meteor) repository locally to test. I verified that analysis results were identical with & without batching. I got a rough sense of memory consumption by watching `top` in my docker VM while analysis ran, and the batching seemed to bring down peak memory usage by ~500mb, which is nothing to sneeze at.

Worth mentioning I did try this *without* GC-forcing first, and memory usage was basically unchanged from `master`. I'm not sure whether blame for this lies with node itself or something eslint does internally, but something is a bit strange with the default GC behavior, I think.

cc @codeclimate/review @jp @GordonDiggs

### DIFF
--- a/bin/node_gc
+++ b/bin/node_gc
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+/usr/bin/env node --expose-gc "$@"
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -1,10 +1,10 @@
-#!/usr/bin/env node
+#!/usr/src/app/bin/node_gc
 
 var CLIEngine = require("eslint").CLIEngine;
 var fs = require("fs");
 var glob = require("glob");
 var options = { extensions: [".js"], ignore: true, reset: false, useEslintrc: true };
-var cli = runWithTiming("cliInit", function() { return new CLIEngine(options); });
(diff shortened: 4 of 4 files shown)


bin/node_gc
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
```
pbrisbin: :lipstick: extra whitespace (i think?)

bin/node_gc
```diff
@@ -0,0 +1,3 @@
...
+/usr/bin/env node --expose-gc $@
```
pbrisbin: Missing quotes: `"$@"`

pbrisbin:
LGTM -- can the same pattern be applied in the other node-based engines?

bin/node_gc
wfleming: Don't we normally put a blank line after the shebang?

wfleming:
@pbrisbin re other node-based engines: I see no reason it couldn't be in cases where we see similar memory problems. It kind of does have to be case-by-case, though. E.g. a theoretical node engine that did cross-file analysis (a la Brakeman) could be broken by this technique, unless we did work to add options to the core analysis tool first.

bin/node_gc
pbrisbin: I know I don't, I don't know what *we* do :)

wfleming: Well, *I* do, so until we have a style guide for shell scripts, I'm leaving it. :stuck_out_tongue_closed_eyes:

mrb:
Very dope @wfleming

PR #38, opened 2015-10-16
Change directory to /code before running analysis.

GordonDiggs:
Some ESLint functions (namely cli.isPathIgnored) depend on relative
paths, so if we aren't cd'd into /code, it can't find the .eslintignore
file correctly.

@codeclimate/review @mrb


fhwang:
LGTM

PR #40, opened 2015-10-20
Prune files within symlinked paths

dblandin:
Despite documentation from `node-glob` which states that symlinked
directories are not followed and crawled, that doesn't seem to be the
case, causing errors when listing files on CC or validating issues
emitted from ESLint regarding ignored files. This adds behavior to
manually prune any path within a symlink before analysis begins.

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -32,7 +32,9 @@ function buildIssueJson(message, path) {
     categories: ["Style"],
     check_name: checkName,
     description: message.message,
-    content: contentBody(checkName),
+    content: {
+      body: contentBody(checkName)
+    },
     location: {
       path: path,
       positions: {


bin/eslint.js
```diff
@@ -56,18 +56,17 @@ function contentBody(check) {
...
     stats.isFile() &&
-    !stats.isSymbolicLink()
-    && extensions.indexOf(extension) >= 0
+    !stats.isSymbolicLink() &&
+    extensions.indexOf(extension) >= 0
   );
 }
 
-function isFileIgnoredByLibrary(file) {
-  var path = file.replace(/^\/code\//, "");
```
GordonDiggs: I found without this `replace`, the call to `isPathIgnored` did not return correctly. Can you confirm that this still works?

dblandin: @GordonDiggs Just simplified this PR without the drive-by refactoring.

dblandin: One of the file strategies has a hard-coded `/code/` prefix which likely relies on that replace for the ignore logic to behave as intended.

bin/eslint.js
```diff
@@ -76,6 +76,18 @@ function isFileIgnoredByLibrary(file) {
   return ignored;
 }
 
+function prunePathsWithinSymlinks(paths) {
```
GordonDiggs: Can you explain what this function is doing?

dblandin: Sure, this is extracting all of the symlinks from `paths` and filtering out any paths which contain the symlink at the front of the path.

For example, if we have
```
a/       -> not a symlink
a/a.txt  -> not a symlink
b/       -> symlink
b/b.txt  -> not a symlink
```

we'll be left with
```js
["a/", "a/a.txt"]
```

GordonDiggs: Cool - might be good to add a comment to that effect. WDYT?

GordonDiggs:
LGTM

PR #41, opened 2015-10-20
Fix content format

GordonDiggs:
The spec requires this to be a hash with a "body" key

@codeclimate/review


pbrisbin:
Nice catch. LGTM

PR #42, opened 2015-10-20
Prune files within symlinked paths

dblandin:
_Follows #40 (accidentally closed due to empty branch push)_

Despite documentation from `node-glob` which states that symlinked
directories are not followed and crawled, that doesn't seem to be the
case, causing errors when listing files on CC or validating issues
emitted from ESLint regarding ignored files. This adds behavior to
manually prune any path within a symlink before analysis begins.

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -76,6 +76,23 @@ function isFileIgnoredByLibrary(file) {
   return ignored;
 }
 
+function prunePathsWithinSymlinks(paths) {
+  // Extracts symlinked paths and filters them out, including any child paths
+  var symlinks = paths.filter(function(path) {
+    return fs.lstatSync(path).isSymbolicLink();
+  });
+
+  return paths.filter(function(path) {
+    var withinSymlink = false;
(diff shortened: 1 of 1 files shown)


dblandin:
@GordonDiggs Would you mind giving this one more look?

wfleming:
I'm late, but have a question: are we relatively sure this issue doesn't affect other engines? And does this problem indicate that we should think about moving this logic into `codeclimate`? I.e. Should codeclimate provide a fully vetted list of individual files & not a mix of files & directories.

fhwang:
Yeah, I think my vote would be to fix it upstream. The odds of this happening in another engine are fairly high I'd say.

dblandin:
I agree, especially when we know that our web interface has trouble displaying details for files within symlinked paths. I'm glad that this fix is in place, as the issue was causing a lot of user-facing problems, but we should definitely consider moving this upstream.

Do you think the move is worth scheduling now or should I investigate whether this issue in fact does affect other engines?

PR #44, opened 2015-10-26
Categories / Remediation points

GordonDiggs:
@codeclimate/review 

Each commit is reviewable separately.

### DIFF
--- a/lib/checks.js
+++ b/lib/checks.js
@@ -27,6 +27,29 @@ var categories = function(checkName) {
   return [checkCategoryMapping[checkName] || "Style"]
 };
 
+var remediationPoints = function(checkName, message) {
+  if (checkName === "complexity") {
+    // (@base_cost + (overage * @cost_per))*1_000_000
+    // cost_per: 0.1,  base: 1
+    var costPer = 100000,
+        points = 1000000,
+        threshold = 10, // TODO: Get this from the eslintrc
+        overage, complexity;
+
+    complexity = message.message.match(/complexity of (\d+)/)[1];
+    overage = complexity - threshold;
+
+    if (overage > 0) {
+      points += (costPer * overage);
+    }
+
+    return points;
+  } else {
+    return 50000;
+  }
+};
+
 module.exports = {
-  categories: categories
+  categories: categories,
+  remediationPoints: remediationPoints
 };
--- a/lib/checks.js
+++ b/lib/checks.js
@@ -1,0 +1,32 @@
+var checkCategoryMapping = {
+  "complexity": "Complexity",
+  "no-unused-vars": "Bug Risk",
+  "no-empty": "Bug Risk",
+  "no-console": "Bug Risk",
+  "no-constant-condition": "Bug Risk",
+  "no-debugger": "Bug Risk",
+  "no-delete-var": "Bug Risk",
+  "no-dupe-keys": "Bug Risk",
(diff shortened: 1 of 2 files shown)


GordonDiggs:
Sample issue with this:

```json
{
  "type":"issue",
  "categories":[
    "Complexity"
  ],
  "check_name":"complexity",
  "description":"Function 'veryComplexFunction' has a complexity of 17.",
  "content":{
    "body":"# Limit Cyclomatic Complexity (complexity)\n\nCyclomatic complexity measures the number of linearly independent paths through a program's source code. This rule allows setting a cyclomatic complexity threshold.\n\n```js\nfunction a(x) {\n    if (true) {\n        return x; // 1st path\n    } else if (false) {\n        return x+1; // 2nd path\n    } else {\n        return 4; // 3rd path\n    }\n}\n```\n\n## Rule Details\n\nThis rule is aimed at reducing code complexity by capping the amount of cyclomatic complexity allowed in a program. As such, it will warn when the cyclomatic complexity crosses the configured threshold.\n\nThe following patterns are considered problems:\n\n```js\n/*eslint complexity: [2, 2]*/\n\nfunction a(x) {               /*error Function 'a' has a complexity of 3.*/\n    if (true) {\n        return x;\n    } else if (false) {\n        return x+1;\n    } else {\n        return 4; // 3rd path\n    }\n}\n```\n\nThe following patterns are not considered problems:\n\n```js\n/*eslint complexity: [2, 2]*/\n\nfunction a(x) {\n    if (true) {\n        return x;\n    } else {\n        return 4;\n    }\n}\n```\n\n## When Not to Use It\n\nIf you can't determine an appropriate complexity limit for your code, then it's best to disable this rule.\n\n## Further Reading\n\n* [About Complexity](http://jscomplexity.org/complexity)\n* [Complexity Analysis of JavaScript Code](http://ariya.ofilabs.com/2012/12/complexity-analysis-of-javascript-code.html)\n\n## Related Rules\n\n* [max-depth](max-depth.md)\n* [max-len](max-len.md)\n* [max-nested-callbacks](max-nested-callbacks.md)\n* [max-params](max-params.md)\n* [max-statements](max-statements.md)\nSource: http://eslint.org/docs/rules/\n"
  },
  "location":{
    "path":"lib/client/init.js",
    "positions":{
      "begin":{
        "line":23,
        "column":1
      },
      "end":{
        "line":23,
        "column":1
      }
    }
  },
  "remediation_points":1700000,
  "engine_name":"eslint"
}
```

lib/checks.js
```diff
@@ -0,0 +1,55 @@
...
+  "no-obj-calls": "Bug Risk",
+  "no-octal": "Compatibility",
+  "no-undef": "Bug Risk",
+  "no-unreachable": "Bug Risk",
+  "use-isnan": "Bug Risk",
+  "valid-typeof": "Bug Risk"
+};
+
+var categories = function(checkName) {
+  return [checkCategoryMapping[checkName] || "Style"]
```
ABaldwinHunter: nice. I like this check

ABaldwinHunter: are the outer brackets necessary?

GordonDiggs: Yep - `categories` is an array

ABaldwinHunter:
@GordonDiggs Nice work! I like this implementation. 

What do you think about adding a spec?

PR #45, opened 2015-10-27
Update react plugin

GordonDiggs:
@codeclimate/review 

Diff: https://github.com/yannickcr/eslint-plugin-react/compare/v3.5.1...v3.6.3


pbrisbin:
LGTM, worth throwing in a compare URL or too noisy?

GordonDiggs:
@pbrisbin done - good idea

pbrisbin:
Cool. Not worth going back, but probably best to ammend those into the commit messages too. PRs are harder to track down after the fact IMO.

PR #58, opened 2016-01-12
Use process.stdout.write instead of console.log and increase batch size

GordonDiggs:
`console.log` does some extra work to construct the output, and because
it's already a string, we can save some memory by using
`process.stdout.write`. See this issue for more information:
https://github.com/nodejs/node/issues/1741#issuecomment-103827443

This is an attempt to remedy the OOM issues when rendering all the
issues. Increasing the batch size will also help by reducing the number
of GCs.

@codeclimate/review

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -194,7 +194,7 @@ var analysisFiles = runWithTiming("buildFileList", function() {
 
 function analyzeFiles() {
   var batchNum = 0
-    , batchSize = 1
+    , batchSize = 10
     , batchFiles
     , batchReport;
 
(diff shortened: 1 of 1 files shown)


wfleming:
I think it's hilarious our `batchSize` was 1 for a while. I think it used to be rather larger than 10, actually, with good results? Probably something to look at separately.

LGTM. Do you know if `write` also flushes the buffer or are you deciding not to force that for now?

GordonDiggs:
@wfleming it used to be 100, which caused issues (I forget exactly what they were though).

GordonDiggs:
In terms of flushing the buffer, I was wrong about what was going on. See the linked github issue for what is _actually_ happening.

PR #60, opened 2016-01-12
Complexity Updates

wfleming:
* Qualify `max-statements` (long methods) as complexity
* Update remediation points calculation to reflect RuboCop base point/per-violation values
* Pull threshold values from actual config for the purposes of calculating remediation points.

:eyeglasses: @codeclimate/review


wfleming:
Note the changed eslint config turns up a complexity complaint in the `bin/eslint` code: i'm inclined to leave that as is for now.

lib/checks.js
```diff
@@ -113,21 +114,48 @@ var categories = function(checkName) {
...
 
-var remediationPoints = function(checkName, message) {
-  if (checkName === "complexity") {
+// Here Be Dragons: this function extracts the relevant value that triggered the issue for
+// checks in the Complexity category. Unfortunately, these values are not available in a
+// structured way, so we extract them from strings. That means that any check categorized
+// as Complexity MUST have a rule here to extract value.
+//
+// If a matching string segment cannot be found, `null` will be returned.
+var messageMetricValue = function(message) {
```
GordonDiggs: Could you add some tests for this function?

lib/checks.js
```diff
@@ -113,21 +114,48 @@ var categories = function(checkName) {
...
+    return parseInt(match[1], 10);
+  }
+  return null;
+};
+
+var metricThreshold = function(ruleId, eslintConfig) {
+   return eslintConfig.rules[ruleId][1];
+};
+
+var remediationPoints = function(checkName, message, eslintConfig) {
```
GordonDiggs: Could you add some tests here too?

wfleming: On the one hand: this means finally adding some tests for this engine, which makes me happy. On the other hand, this means I'm responsible for setting up the testing setup on this engine, which makes me sad.

But I agree it's probably worth testing stuff here. I'll take a crack at it.

wfleming:
@GordonDiggs tests have been added if you'd like to take a look!

@ABaldwinHunter this seems relevant to your recent interests (viz. your lunch and learn talk), so it'd be helpful if you took a look as well.

Makefile
```diff
@@ -1,6 +1,9 @@
-.PHONY: image
+.PHONY: image test
 
 IMAGE_NAME ?= codeclimate/codeclimate-eslint
 
 image:
 	docker build --rm -t $(IMAGE_NAME) .
+
+test: image
+	docker run $(IMAGE_NAME) node_modules/.bin/mocha
```
GordonDiggs: To be a bit more conventional, can you specify a task inside the `package.json` file? Something like:

```json
"scripts": {
    "test": "mocha test"
  }
```

And then this invocation can be `docker run $(IMAGE_NAME) npm run test`

GordonDiggs:
One comment but otherwise this looks great! Excited to actually have tests here now.

Makefile
wfleming: Good suggestion: done.

GordonDiggs:
:+1:

wfleming:
As noted above, the CC issue is a function complexity thing in `bin/eslint` that's been there a while: I think that's out of scope to fix now, so I'm going to mercy merge.

PR #61, opened 2016-01-13
Dockerfile: use tag for base image, add maintainer

wfleming:
Both of these are small best practice things.

1. Use a tag on our base image to reduce chance of random breakage
2. Add maintainer info

:eyes: @codeclimate/review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,18 @@
-FROM mhart/alpine-node
+FROM mhart/alpine-node:5.4
+MAINTAINER Code Climate <hello@codeclimate.com>
 
 WORKDIR /usr/src/app
 COPY package.json /usr/src/app/
 
-RUN apk --update add git
+RUN apk --update add git && \
+    npm install && \
+    apk del --purge git
 
-RUN npm install
+COPY . /usr/src/app
 
 RUN adduser -u 9000 -D app
 USER app
-
-COPY . /usr/src/app
+VOLUME /code
+WORKDIR /code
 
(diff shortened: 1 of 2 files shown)


GordonDiggs:
LGTM

pbrisbin:
While you're in here, can you fix up some other things mentioned [here](https://github.com/codeclimate/spec/blob/master/SPEC.md#packaging)

Ones I frequently notice being missed, and are missed here:

- The /code must be declared as a VOLUME.
- The WORKDIR must be specified as /code

wfleming:
Done @pbrisbin 

Switching to the tagged base image actually revealed a problem as well: I think newer node releases are back to needing to install npm separately rather than having come with `node`? I'll look into that in a bit.

wfleming:
Updated and improved, @codeclimate/review

Dockerfile
```diff
@@ -7,9 +8,14 @@ RUN apk --update add git
 
 RUN npm install
 
-RUN adduser -u 9000 -D app
-USER app
+
+RUN apk del --purge git
```
pbrisbin: If the intention here is to keep image size down, this won't do it since it's in a separate layer. Something like:

```
RUN apk --update add git && npm install && apk del --purge git
```

Would do that though. If the intent is just to keep git out of the final image, then it doesn' t matter.

wfleming: Interesting: I didn't realize that. Makes sense when you think it through, though. Will update.

wfleming: @pbrisbin updated

pbrisbin:
LGTM

PR #62, opened 2016-01-13
Default to using babel for parsing

wfleming:
The version of esprima/espree that eslint defaults to using for parsing
still doesn't work super well for ES6. (Details at
http://eslint.org/blog/2014/12/espree-esprima/).

This changes options to default to using babel for parsing. Also
updating the babel related packages here as well.

If any project does not want babel, it can still be overridden in
`.eslintrc` with a `parser` setting.

:eyes: @codeclimate/review


GordonDiggs:
LGTM. For consistency, should we enable this in our default `eslintrc` as well?

wfleming:
@GordonDiggs That's an interesting suggestion, but I think it's an either-or kind of thing. Either we should put it in the default `.eslintrc`, OR we should just default to it here. Any setting in the `.eslintrc` would override this, anyway.

The motivation for doing it here was that @noahd1 was confused about why he wasn't getting results on the `relax` repo. He had configured an `.eslintrc` there, but didn't specify the parser. So I guess the argument for doing it here is that it's probably not going to be all that uncommon for projects to show up with pre-established `.eslintrc` that don't specify a parser because the pre-existed ES6 and are transitioning, or were configured using `eslint` instead of `codeclimate init`, or whatever. If we can work out of the box for those cases, great.

Does that make sense? I'm curious if others on the team think it's worth doing both or would prefer to pick one approach & stick to it.

GordonDiggs:
@wfleming What you said makes sense to me. Given that I don't see any good advantage of esprima over babel, setting it here as the default makes sense.

noahd1:
@wfleming Now that I see @GordonDiggs question I'm a bit more on the fence.. Keep in mind I'm not the target audience not being a day to day javascript developer. 

This creates a bit of a surprising situation where our default _runtime_ behavior differs from eslint's _runtime_ behavior which may lead to some head on table confusion down the line if you were to compare the output side by side.

noahd1:
Is there a conflict here between transparency on how Code Climate works (no magic) and ease of use (good defaults)? Is there a way to get both?

wfleming:
@noahd1 

> This creates a bit of a surprising situation where our default runtime behavior differs from eslint's runtime behavior which may lead to some head on table confusion down the line if you were to compare the output side by side.

I don't think there should be output differences, excepting a case where babel had a parsing bug that espree didn't. My understanding was that the point of the `babel-eslint` plugin was to ensure it would give ESLint an equivalent AST for it to inspect.

> Is there a conflict here between transparency on how Code Climate works (no magic) and ease of use (good defaults)? Is there a way to get both?

I'd argue that since this codebase is open source this is already non-magical to the extent any of our engines are non-magical. This is a choice that provides "good defaults" for the most cases while still being user-adjustable, I think.

We also took a similar path with radon recently, FWIW: that engine defaults to using python3, but can be configured to use python2. We don't autogenerate any explicit config for it either way, we just cover it in the docs.

I'll let this bake, for now: I can pick it up next week, or somebody else can push it to completion in the next couple days when we decide the right way to handle this.

noahd1:
Radon is an interesting counter example. I think what we're seeing is that
some tools offer pluggable parsers and some don't. Or some use "hosted"
parsers and some don't.

With radon there's an understanding, I think, that the parser is based on
the version of Python you're running, aka your environment and is therefore
not the tool's concern. So for us, we are forced to pick a default
environment/parser/Python version, the tool has no opinion on it.

With eslint it seems that it's pluggable and environment doesn't matter.
The tool _does_ define a default, and we're actually choosing something
other than the default, using something other than a configuration file
where these overrides are usually defined.

Or maybe I'm over-thinking this. Curious if @brynary has an opinion on this.

On Wed, Jan 13, 2016 at 7:23 PM Will Fleming <notifications@github.com>
wrote:

> @noahd1 <https://github.com/noahd1>
>
> This creates a bit of a surprising situation where our default runtime
> behavior differs from eslint's runtime behavior which may lead to some head
> on table confusion down the line if you were to compare the output side by
> side.
>
> I don't think there should be output differences, excepting a case where
> babel had a parsing bug that espree didn't. My understanding was that the
> point of the babel-eslint plugin was to ensure it would give ESLint an
> equivalent AST for it to inspect.
>
> Is there a conflict here between transparency on how Code Climate works
> (no magic) and ease of use (good defaults)? Is there a way to get both?
>
> I'd argue that since this codebase is open source this is already
> non-magical to the extent any of our engines are non-magical. This is a
> choice that provides "good defaults" for the most cases while still being
> user-adjustable, I think.
>
> We also took a similar path with radon recently, FWIW: that engine
> defaults to using python3, but can be configured to use python2. We don't
> autogenerate any explicit config for it either way, we just cover it in the
> docs.
>
> I'll let this bake, for now: I can pick it up next week, or somebody else
> can push it to completion in the next couple days when we decide the right
> way to handle this.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/codeclimate/codeclimate-eslint/pull/62#issuecomment-171482421>
> .
>
-- 
Noah Davis, VP of Operations / noah@codeclimate.com / We're hiring! /
http://jobs.codeclimate.com

jpignata:
Assuming that babel and espree parse "vanilla" JavaScript in generally the same way and ESLint produces identical issues between the two of them, it seems like it would be preferable to have better ES6 support out of the box.

We might be overthinking this one – it should be an invisible change save for better results for ES6.

noahd1:
Sold

On Wed, Jan 13, 2016 at 8:29 PM John Pignata <notifications@github.com>
wrote:

> Assuming that babel and espree parse "vanilla" JavaScript in generally the
> same way and ESLint produces identical issues between the two of them, it
> seems like it would be preferable to have better ES6 support out of the box.
>
> We might be overthinking this one – it should be an invisible change save
> for better results for ES6.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/codeclimate/codeclimate-eslint/pull/62#issuecomment-171493816>
> .
>
-- 
Noah Davis, VP of Operations / noah@codeclimate.com / We're hiring! /
http://jobs.codeclimate.com

PR #63, opened 2016-01-25
Revert babel as the default parser.

wfleming:
This reverts commit df26467d17d067cde41760ac4c314d8a04241362, reversing
changes made to 804d90fdbb99bce9d218df7993bfec65a913b02e.

There are cases that cause the stack to blow up with Babel, apparently: https://github.com/eslint/eslint/issues/3440

These issues seem to be fixed in eslint's master, but they're currently
in beta of a 2.0 release, so trying to merge in that with our changes
seems like a riskier change to make right now: we should wait until
they release 2.0 & take another look. For now, if you're using ES6, you
can opt into this config

cc @codeclimate/review


GordonDiggs:
LGTM

PR #64, opened 2016-01-25
Default line/column when null

wfleming:
"fatal" issues that eslint reports might not have column/linenumbers.
This is invalid by the engine spec, so we should avoid it.

@codeclimate/review


pbrisbin:
LGTM

PR #67, opened 2016-02-03
Update README.md

leftees:
Add configuration subheading, explaining extensions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,18 @@ ESLint is a tool for identifying and reporting on patterns found in ECMAScript/J
 2. Run `codeclimate engines:enable eslint`. This command both installs the engine and enables it in your `.codeclimate.yml` file.
 3. You're ready to analyze! Browse into your project's folder and run `codeclimate analyze`.
 
+### Configuration
+
+You can configure our ESLint Engine to analyze the file types you’d like by configuring the extensions in your `.codeclimate.yml`, as shown in the example below:
+
+```eslint:
+   enabled: true
+   config:
+    extensions:
+   - .es6
+   - .js
+   - .jsx```
+
 ### Need help?
 
 For help with ESLint, [check out their documentation](https://github.com/eslint/eslint).


README.md
```diff
@@ -12,6 +12,18 @@ ESLint is a tool for identifying and reporting on patterns found in ECMAScript/J
 2. Run `codeclimate engines:enable eslint`. This command both installs the engine and enables it in your `.codeclimate.yml` file.
 3. You're ready to analyze! Browse into your project's folder and run `codeclimate analyze`.
 
+### Configuration
+
+You can configure our ESLint Engine to analyze the file types you’d like by configuring the extensions in your `.codeclimate.yml`, as shown in the example below:
+
+```eslint:
```
wfleming: Formatting on this block is off: first line of content should start on a new line after the <code>```</code>, sub-values should be indented 2 spaces instead of 1, and the extension lines starting with `-` should line up with `extensions`. Like:

<pre>
```yaml
eslint:
  enabled: true
  config:
    extensions:
    - .es6
    - .js
    - .jsx
```
</pre>

wfleming:
Has there been discussion of to what degree these README's should cover the same content as the [docs articles](https://docs.codeclimate.com/docs/eslint), or whether one should generally just link to the other? Trying to keep the same content up to date in both places is going to get annoying, and error-prone.

IMHO, if we were going to prefer one, we should probably prefer docs in-repo, since I think most devs nowadays are conditioned to go straight to a tool's github repo for these kinds of details.

README.md
```diff
@@ -12,6 +12,18 @@ ESLint is a tool for identifying and reporting on patterns found in ECMAScript/J
 2. Run `codeclimate engines:enable eslint`. This command both installs the engine and enables it in your `.codeclimate.yml` file.
 3. You're ready to analyze! Browse into your project's folder and run `codeclimate analyze`.
 
+### Configuration
+
+You can configure our ESLint Engine to analyze the file types you’d like by configuring the extensions in your `.codeclimate.yml`, as shown in the example below:
```
wfleming: Perhaps we should note the default value here so users can more easily know whether they need to customize this. Our default is to only analyze `".js"`.

wfleming:
@leftees This PR is a bit stale. Do you want to update it, or is it no longer as relevant?

dblandin:
Going to go ahead and close this PR for now.

PR #69, opened 2016-02-08
Analyze files with .es6 extensions by default

ABaldwinHunter:
We should also analyze .es6 file extensions by default.

This creates an easier setup for users with `.es6` files

I don't believe there's anything special about the syntax to be avoided, but glad to be corrected if not the case.

cc @codeclimate/review


GordonDiggs:
Will es6 parse correctly without the babel parser? We defaulted to babel for a while but backed off of that due to errors

ABaldwinHunter:
@GordonDiggs Ah okay, maybe not. Looks like Esprima is [starting to support es6](http://esprima.org/doc/es6.html) and has an [experimental Harmony branch](https://github.com/addyosmani/es6-tools#parsers) but perhaps is not all the way there. 

I'll go ahead and close this PR for now. Can revisit later. Thanks for review.

PR #71, opened 2016-02-09
Only output timing information if debug is set to true.

GordonDiggs:
Trying to make STDERR quieter

@codeclimate/review


ABaldwinHunter:
Curious about motivation and whether there's a different spot that STDERR is getting in the way?

Up to you, but maybe consider adding a note about motivation to commit message?

One optional nit, otherwise LGTM.

wfleming:
838f269 seems super important: that could have something to do with the "invalid output" issues we see from eslint regularly.

PR #72, opened 2016-02-10
Update eslint-plugin-react

jakemhiller:
This resolves issues with missing rules that are used in the `airbnb` preset.

I'm currently seeing errors like this:

`Definition for rule 'react/no-deprecated' was not found`

This is because airbnb is referencing rules that aren't defined in this version of the react plugin.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "eslint": "codeclimate/eslint.git#d24d0b",
     "eslint-config-airbnb": "^5.0.0",
     "eslint-plugin-babel": "2.1.1",
-    "eslint-plugin-react": "3.6.3",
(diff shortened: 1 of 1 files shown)


jakemhiller:
For anyone else that has this issue and needs a fix, just change:

`"extends": "airbnb"` to `"extends": "airbnb/base"`

and if you still want to use the react plugin, make sure at least this rule is set:

```js
{
    "react/jsx-uses-react": 1
}
```

This gets rid of the `React is not defined` errors when using jsx.

wfleming:
Thanks @jakemhiller!

jakemhiller:
Thanks!

PR #73, opened 2016-02-12
Error if there is not a valid configuration file

GordonDiggs:
If there is no eslintrc.* file, ESLint's behavior is to do nothing (and
therefore emit no issues). Because this can be confusing, we should fail
hard and show helpful information.

screen shot 2016-02-12 at 10 54 48 am

@codeclimate/review

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -15,8 +15,7 @@ var options = { extensions: [".js"], ignore: true, reset: false, useEslintrc: tr
 var cli = new CLIEngine(options);
 var debug = false;
 var checks = require("../lib/checks");
-
-console.error("ESLint is running with the " + cli.getConfigForFile(null).parser + " parser.");
+var validateConfig = require("../lib/validate_config");
 
 // a wrapper for emitting perf timing
 function runWithTiming(name, fn) {
@@ -232,4 +231,12 @@ function analyzeFiles() {
   }
 }
 
-analyzeFiles();
+if (validateConfig(options.configFile)) {
+  console.error("ESLint is running with the " + cli.getConfigForFile(null).parser + " parser.");
+
+  analyzeFiles();
+} else {
+  console.error("No rules are configured. Make sure you have added a config file with rules enabled.");
+  console.error("See our documentation at https://docs.codeclimate.com/docs/eslint for more information.");
+  process.exit(1);
+}
--- a/lib/validate_config.js
+++ b/lib/validate_config.js
@@ -1,0 +1,13 @@
+var CLIEngine = require("eslint").CLIEngine
+  , cli = new CLIEngine()
+  , fs = require("fs");
+
+module.exports = function(configPath) {
+  if (configPath) {
+    return true;
+  } else {
+    var config = cli.getConfigForFile(null);
+
+    return Object.keys(config.rules).length > 0;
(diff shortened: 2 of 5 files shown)


bin/eslint.js
```diff
@@ -232,4 +231,12 @@ function analyzeFiles() {
...
 }
 
-analyzeFiles();
+if (validateConfig(options.configFile)) {
+  console.error("ESLint is running with the " + cli.getConfigForFile(null).parser + " parser.");
+
+  analyzeFiles();
+} else {
+  console.error("No rules are configured. Make sure you have added a config file with rules enabled.");
+  console.error("See our documentation at docs.codeclimate.com/docs/eslint for more information.");
```
wfleming: This is a nit, but can we put a protocol on the URL?

ABaldwinHunter: What do you think about specifying that it's a `.eslintrc` or `.eslintrc.js` config file?

test/validate_config_test.js
```diff
@@ -0,0 +1,46 @@
...
+      if (err) throw err;
+
+      process.chdir(directory);
+
+      var configPath = path.join(directory, ".eslintrc.json");
+      var config = {
+        rules: {
+          strict: 0
+        }
+      };
```
ABaldwinHunter: nice tests!

bin/eslint.js
GordonDiggs: @wfleming yep sounds good

@ABaldwinHunter I think there are too many possibilities to list here (`.eslintrc(.js|.json|.yml.yaml|nothing`) and it would be a bit wordy. My link to the docs was so that people would see it there

ABaldwinHunter: Okay sounds good :+1:

ABaldwinHunter:
@GordonDiggs one optional comment but LGTM!
love the specs.

Unrelated to shipping this change, but wdyt about adding having this validation tripped in `codeclimate validate-config`?

ABaldwinHunter:
Might be outside the purview of validating the `.codeclimate.yml`.

wfleming:
> Unrelated to shipping this change, but wdyt about adding having this validation tripped in codeclimate validate-config?

Personally I think the CLI already knows too much about specific engines. This is an eslint behavior, not a general engine behavior, so it doesn't seem appropriate to me for the CLI to do this.

GordonDiggs:
I agree with @wfleming. Additionally, in the ideal workflow, a user will call `codeclimate init` first and never see this.

lib/validate_config.js
```diff
@@ -0,0 +1,13 @@
+var CLIEngine = require("eslint").CLIEngine
+  , cli = new CLIEngine()
+  , fs = require("fs");
+
+module.exports = function(configPath) {
+  if (configPath) {
```
wfleming: This behavior is a little surprising to me: this would mean if you explicitly told us to use `.eslintrc.foo` as a config file, but that file didn't exist, we would still pass that as valid?

GordonDiggs: @wfleming yeah - if you've specified a specific config file this check would pass, but a later one would error.

wfleming: Gotcha. Makes sense.

PR #83, opened 2016-03-15
Remove `exclude_paths` support

wfleming:
We no longer provide these in any config visible to engines.

@codeclimate/review

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -105,30 +105,6 @@ function prunePathsWithinSymlinks(paths) {
   });
 }
 
-function exclusionBasedFileListBuilder(excludePaths) {
-  // Uses glob to traverse code directory and find files to analyze,
-  // excluding files passed in with by CLI config, and including only
-  // files in the list of desired extensions.
-  //
-  // Deprecated style of file expansion, supported for users of the old CLI.
-  return function(extensions) {
-    var analysisFiles = [];
-    var allFiles = glob.sync("/code/**/**", {});
-
-    prunePathsWithinSymlinks(allFiles).forEach(function(file, i, a){
-      if(excludePaths.indexOf(file.split("/code/")[1]) < 0) {
-        if(fs.lstatSync(file).isFile()) {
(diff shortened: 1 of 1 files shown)


bin/eslint.js
```diff
@@ -168,12 +145,9 @@ runWithTiming("engineConfig", function () {
...
         engineConfig.include_paths
       );
-    } else if (engineConfig.exclude_paths) {
-      var ignores = engineConfig.exclude_paths;
-      buildFileList = exclusionBasedFileListBuilder(ignores);
     } else {
-      // No includes or excludes, let's try with everything
-      buildFileList = exclusionBasedFileListBuilder([]);
+      // No explicit includes, let's try with everything
+      buildFileList = inclusionBasedFileListBuilder(["./"]);
```
ABaldwinHunter: Not sure if analyzing all files in absence of include_paths is the desired behavior? Would there be any cases where no include paths get passed, e.g. calling `analyze` on an excluded directory?

ABaldwinHunter:
One question, otherwise LGTM! Nice to see the excludes go

bin/eslint.js
```diff
@@ -105,31 +105,8 @@ function prunePathsWithinSymlinks(paths) {
...
-        }
-      }
-    });
-
-    return analysisFiles;
-  };
-}
-
 function inclusionBasedFileListBuilder(includePaths) {
+  process.stderr.write("inclusion file builder with include paths: " + includePaths + "\n");
```
GordonDiggs: should we wrap this in an `if (debug)` check?

wfleming: Yes we should.

bin/eslint.js
wfleming: This has been the default behavior in many of our engines, including this one, for a while: there are no production instances where it should occur, but it continues to seem like a reasonable default behavior in the absence of anything explicit.

bin/eslint.js
wfleming: Opted to just remove it, actually. Not sure it's useful longer term.

PR #84, opened 2016-03-15
Keep paths relative to /code

wfleming:
We saw a bug with `.eslintignore`: when we tested whether a file was
being ignored, we passed the relative directory (which is correct, and
produces expected behavior), but when we actually passed the file for
analyiss, we were passing the absolute path. Thus, a file that actually
should not have been ignored was detected as ignored by eslint when
asked to analyze it.

The changes here keep file paths relative for the most part: we chdir to
/code before we do just about anything else, so this works well. I also
extracted all the remaining usage of "/code" to a global var just for
organizational purposes.

@codeclimate/review

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -1,6 +1,7 @@
 #!/usr/src/app/bin/node_gc
 
-process.chdir('/code');
+var CODE_DIR = "/code";
+process.chdir(CODE_DIR);
 
 // Redirect `console.log` so that we are the only ones
 // writing to STDOUT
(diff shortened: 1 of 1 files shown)


GordonDiggs:
LGTM - are there any tests we could write for this?

wfleming:
> are there any tests we could write for this

We're not well set up to test the integration-y things like this for this engine at the moment, unfortunately. I did run it across all the js repos I have locally to look for regressions.

I've considered several times breaking down this code into more modules that could be individually & more easily tested, but since this is likely to require a near-total rewrite for eslint 2.0, I've kept deciding not to do it until we need to do it for that. Does that seem reasonable to you?

ntwb:
> ...this is likely to require a near-total rewrite for eslint 2.0

Oh! @wfleming I've started watching this repo as I thought it would be a good one to familiarise myself with as I'd like to create a Code Climate Engine for [stylelint](https://github.com/stylelint/stylelint), any tips on a repo that I should watch alternatively that is in a more modulised code format as your envisaging above?

Issue #82, opened 2016-03-11
eslint config file path is ignored.


unageanu:
Hi.
I specified a eslint config file path within my `.codeclimate.yml`, but it ignored.
My .codeclimate.yml looks like below:

```
engines:
  eslint:
    
    enabled: true

    config:
      config: sites/config/.eslintrc
```

I think it got worse when a creation of a CLIEngine ( `new CLIEngine(options)` ) moved to the top of the file by [this commit](https://github.com/codeclimate/codeclimate-eslint/commit/fbfb61a834be0f70c890338d26354e87f47deb4d). It seems that the creation must be after [the processing of options]( https://github.com/codeclimate/codeclimate-eslint/blob/master/bin/eslint.js#L163).
What do you think about it?

pbrisbin:
Hi @unageanu ,

I think the issue is the doubly-nested `config` in your yaml. Can you try:

```yaml
engines:
  eslint:
    enabled: true
    config: sites/config/.eslintrc
```

?

Also, if you run `CODECLIMATE_DEBUG=1 codeclimate analyze` you should find a debug-level log entry showing the exact contents of the `/config.json` being passed to the engine.

unageanu:
@pbrisbin, Thanks for reply.
I tried your adviced setting,

```yaml
engines:
  eslint:
    enabled: true
    config: sites/config/.eslintrc
```

But error occurred.

```shell
$ CODECLIMATE_DEBUG=1 /usr/local/bin/codeclimate analyze                                                                                     
Starting analysis
Running eslint: |[DEBUG] engine config: "{\"enabled\":true,\"config\":\"sites/config/.eslintrc\",\"include_paths\":[\".gitignore\",\"Procfile\",\"config.ru\",\"config/\",\"sites/.gitignore\",\"sites/gulpfile.js\",\"sites/package.json\",\"sites/src/\",\"sites/tasks/\",\"sites/cache/\",\"src/\",\"log/\",\"build/\",\"app.json\",\"circle.yml\",\"sample_agents/\",\"Gemfile\",\".codeclimate.yml\",\"README.md\",\"Gemfile.lock\",\"log.txt\"]}"
[...]

PR #85, opened 2016-03-16
Delay instantiation of cli variable

pbrisbin:
Where it was currently instantiated was before the options have been fully
inflated from a possible engine config. This meant that any configured
configFile or extensions would be ignored.

/cc @codeclimate/review

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -13,7 +13,7 @@ var docs = require("eslint").docs;
 var fs = require("fs");
 var glob = require("glob");
 var options = { extensions: [".js"], ignore: true, reset: false, useEslintrc: true };
-var cli = new CLIEngine(options);
+var cli; // instantiation delayed until after options are (potentially) modified
 var debug = false;
 var checks = require("../lib/checks");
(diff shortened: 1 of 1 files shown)


pbrisbin:
Fixes #82

bin/eslint.js
```diff
@@ -159,6 +159,8 @@ runWithTiming("engineConfig", function () {
     if (userConfig.debug) {
       debug = true;
     }
+
+    cli = new CLIEngine(options);
```
wfleming: This is currently inside if the `if config.json exists` block, I think, which I'm not sure is what you want. While the missing config.json shouldn't actually happen, we do normally try to have sane defaults if it's missing, don't we?

I think the simple change, if you agree with that reasoning, is just to move this line down one line, outside of the `if` block.

pbrisbin: My bad, that was entirely unintentional.

pbrisbin: Fixed with another minor change which made sense once I realized `buildFileList` was a similar pattern.

wfleming:
LGTM

PR #88, opened 2016-05-03
Behave like --no-eslintrc if custom config is present

pbrisbin:
The default behavior of eslint --config is to use that config _in addition to_
any default config (e.g. .eslintrc), not _instead of_. To prevent this, you
have to additionally pass --no-eslintrc.

We find this behavior surprising and prefer setting a custom config to also
imply --no-eslintrc, despite being a behavior difference between the engine and
the underlying tool.

/cc @codeclimate/review

This came out of a CX report. I'm open to opinions if departing from the
default behavior of eslint is actually worthwhile. Another option is to somehow
influence --no-eslintrc via codeclimate.yml.


GordonDiggs:
I like this approach. 👍

PR #89, opened 2016-05-05
Add fingerprint override for complexity rule

dblandin:
This PR provides an fingerprint override for the complexity offenses incorporating the path, rule, and function name.

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -17,6 +17,7 @@ var cli; // instantiation delayed until after options are (potentially) modified
 var debug = false;
 var checks = require("../lib/checks");
 var validateConfig = require("../lib/validate_config");
+var computeFingerprint = require("../lib/compute_fingerprint");
 
 // a wrapper for emitting perf timing
 function runWithTiming(name, fn) {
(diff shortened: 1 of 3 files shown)


bin/eslint.js
```diff
@@ -66,6 +67,13 @@ function buildIssueJson(message, path) {
     },
     remediation_points: checks.remediationPoints(checkName, message, cli.getConfigForFile(path))
   };
+
+  var fingerprint = Fingerprint(path, checkName, message.message).compute();
```
pbrisbin: Shouldn't this use `new`?

Alternatively you could just `export` the `compute` function (which seems to be the prevailing pattern in this repo).

```js
var computeFingerprint = require("../lib/compute_fingerprint");

var fingerprint = computeFingerprint(path, checkName, message.message);
```

The implementation could certainly use a constructor and object internally still.

pbrisbin:
The approach LGTM. Just a minor question about the syntax.

bin/eslint.js
dblandin: Good catch and I think switching the export to a function sounds good.

PR #79, opened 2016-02-24
Monkey patch eslint, instead of modifying it

dancrumb:
This isn't ready for merging for the following reasons:

- [x] The skipped test - need clarity around correct behaviour
- [x] Doesn't include the `docs` filed that was created in the codeclimate/eslint fork - I'm currently unsure of what's being done there, so I need to investigate further.

PR #92, opened 2016-05-26
ESlint 2.0

pbrisbin:
This is the branch we'll be deploying the eslint-2 channel from as we begin
initial testing of that image. It's #79 with some additional changes (see
individual commits) and anything we identify and address during testing.

More documentation about this change is on the way, but for anyone who 
sees this PR and wants to use it, the engine can be opted into with:

```yaml
engines:
  eslint:
    enabled: true
    channel: eslint-2
```


chibicode:
@jpignata @dblandin could we add [`eslint-config-standard-jsx`](
https://github.com/feross/eslint-config-standard-jsx) to `package.json` as well? This is required for React developers (and [this is included when you install `standard`](https://github.com/feross/standard/blob/master/eslintrc.json)).

dblandin:
@chibicode That sounds good to me! I'll open up a PR to pull that in.

dblandin:
Hey @chibicode,

I just pushed out a new engine image to the eslint-2 channel with support for the [eslint-config-standard-react] and [eslint-config-standard-jsx] configs.

Let me know if it works for you!

[eslint-config-standard-react]: https://github.com/feross/eslint-config-standard-react
[eslint-config-standard-jsx]: https://github.com/feross/eslint-config-standard-jsx

chibicode:
@dblandin great, thank you!

lib/eslint-patch.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+var meld = require('meld');
+var docs = require('./docs');
+
+var supportedPlugins = ['react', 'babel'];
+
+module.exports = function patcher(eslint) {
+
+  meld.around(eslint.CLIEngine, 'loadPlugins', function(joinPoint) {
```
mhaylock: Is this patch working as intended? I arrived here because I'm seeing `Failed to load plugin` errors for an unsupported plugin, but I _think_ this patch is intended to avoid loading unsupported plugins?

I'm questioning whether this does anything because it appears to me that `eslint.CLIEngine` has not had a `loadPlugins` method since [this commit](https://github.com/eslint/eslint/commit/b28a19d1b708e42bf54c0afc6df363b2c3d250a3), and that commit has been around since ESLint 2.0.0?

pbrisbin: This patch is not working as intended. It was at the time it was written, but (as you noted) that's no longer the case given the current internals of ESLint.

We're evaluating if we want to accept unsupported plugins causing errors like that, or try to bring back the logic intended to replace that with ignoring them. Unfortunately, this `meld` approach doesn't appear viable on the new internals, so it would require moving back to our own fork of ESLint itself.

In the meantime though, we should probably remove this.

dblandin:
Hey everyone!

Development for the `eslint-2` channel has moved to the [channel/eslint-2 branch](https://github.com/codeclimate/codeclimate-eslint/tree/channel/eslint-2). That branch won't have a tracking PR. Instead, new PRs targeting the `eslint-2` channel should be opened with a base of `channel/eslint-2`.

Also, if you're interested, we [just released](https://github.com/codeclimate/codeclimate-eslint/issues/125#issuecomment-246436125) an `eslint-3` channel!

Thanks!

PR #93, opened 2016-05-27
Update dependencies

dblandin:
This updates this engine's npm dependencies for ESLint and our supported ESLint plugins. I've also added an npm shrinkwrap file to lock down the entire dependency tree.

```
eslint@2.4.0               -> eslint@2.10.2
eslint-config-airbnb@6.0.2 -> eslint-config-airbnb@9.0.1
eslint-plugin-babel@2.1.1  -> eslint-config-airbnb@3.2.0
eslint-plugin-react@4.0.0  -> eslint-config-react@5.1.1
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,16 +9,18 @@
   },
   "dependencies": {
     "babel-eslint": "6.0.4",
-    "eslint": "2.4.0",
-    "eslint-config-airbnb": "^6.0.2",
-    "eslint-plugin-babel": "2.1.1",
-    "eslint-plugin-react": "^4.0.0",
-    "glob": "5.0.14",
-    "meld": "^1.3.2"
+    "eslint": "2.10.2",
+    "eslint-config-airbnb": "9.0.1",
+    "eslint-plugin-babel": "3.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -9,16 +9,18 @@
...
+    "eslint-plugin-import": "1.8.1",
+    "eslint-plugin-jsx-a11y": "1.2.2",
+    "eslint-plugin-react": "5.1.1",
+    "glob": "7.0.3",
+    "meld": "1.3.2"
   },
   "devDependencies": {
-    "chai": "^2.1.0",
-    "mocha": "^2.3.4",
+    "chai": "^3.5.0",
```
GordonDiggs: Should these be locked as well?

wfleming:
👍

package.json
dblandin: I think the dev dependencies can be looser here since they're only used for the test suite.

GordonDiggs: Cool - we've seen issues elsewhere with different versions of these being installed, but happy to address that whenever it comes up

PR #95, opened 2016-06-02
Support standard style ESLint shared config and plugin

dblandin:
[Standard Style][] is a commonly adopted style for writing JavaScript. We hope
to support other [third-party plugins][] and [shared configurations][] in the
future without needing to vendor them in but this update removes a blocker for
a bunch of users and I think it's popular enough to warrant pulling into the
engine.

@codeclimate/review :mag_right:

[Standard Style]: http://standardjs.com/
[third-party plugins]: https://github.com/codeclimate/codeclimate-eslint/issues/86
[shared configurations]: https://github.com/codeclimate/codeclimate-eslint/issues/4

PR #96, opened 2016-06-02
eslint-2: Support standard style ESLint shared config and plugin

dblandin:
_This applies a similar update (with different version numbers) from #95 to the `eslint-2` channel._

Standard Style is a commonly adopted style for writing JavaScript. We hope
to support other third-party plugins and shared configurations in the
future without needing to vendor them in but this update removes a blocker for
a bunch of users and I think it's popular enough to warrant pulling into the
engine.

@codeclimate/review 🔎


pbrisbin:
Cool Co-cool.

PR #98, opened 2016-06-02
eslint-2: Update CircleCI config to use patrick 

dblandin:
Applying the same changes here as made in #97.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: image test
+.PHONY: image test citest
 
 IMAGE_NAME ?= codeclimate/codeclimate-eslint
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,9 +1,12 @@
...
+.PHONY: image test citest
 
 IMAGE_NAME ?= codeclimate/codeclimate-eslint
 
 image:
 	docker build --rm -t $(IMAGE_NAME) .
 
 test: image
-	docker run --rm --workdir="/usr/src/app" $(IMAGE_NAME) npm run test
+	docker run --rm $(IMAGE_NAME) sh -c "cd /usr/src/app && npm run test"
```
pbrisbin: Minor, but you can pass `--workdir /usr/src/app` to `docker run` so you don't need to `sh` to `cd`.

PR #107, opened 2016-07-21
Update package.json

leftees:
Add support for `eslint-config-angular` plugin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: image test citest
+.PHONY: image test citest shrinkwrap
 
 IMAGE_NAME ?= codeclimate/codeclimate-eslint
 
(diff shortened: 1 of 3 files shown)


GordonDiggs:
Should there be a change to the shrinkwrap json file too? cc @pbrisbin

pbrisbin:
Yes. As a process though, I think it's reasonable for the engineer to do that as part of picking this up from CX (WDYT?) -- I personally don't know how to update the shrinkwrap file :|

pbrisbin:
@GordonDiggs updated and rebased. Added a `make` task to update the shrinkwrap using the engine image's `npm` too.

GordonDiggs:
Nice LGTM

PR #114, opened 2016-08-11
Heuristically filter out minified files

ABaldwinHunter:
By default we filter out minified files that are identifiable as minified from
their filename (e.g. *.min.js), but many projects have minified files that
don't match those patterns, and these files can cause ESLint to segfault in a
container environment (possibly due to memory pressure).

This adds a new heuristic filter based on line length that filters out files
that appear to be minified to prevent them being analyzed. For any matching
files, a log message is sent to stderr notifying that the file is being
skipped.

@codeclimate/review @wfleming 

Note: tested locally on 8 repos and saw minimal performance loss. There are some minified files that we currently analyze successfully and this change would mean ignoring them instead. We think that is okay.


lib/batch_sanitizer.js
```diff
@@ -0,0 +1,41 @@
...
+
+function BatchSanitizer(files, stderr) {
+  this.files = files;
+  this.stderr = stderr || process.stderr;
+}
+
+BatchSanitizer.prototype.sanitizedFiles = function() {
+  var sanitizedFiles = [];
+
+  for(var i = 0; i < this.files.length; i++) {
```
GordonDiggs: Any reason you did this with a for loop and not something like `forEach`?

lib/batch_sanitizer.js
```diff
@@ -0,0 +1,41 @@
...
+    if (this.isMinified(this.files[i])) {
+      this.stderr.write("WARN: Skipping " + this.files[i] + ": it appears to be minified\n");
```
GordonDiggs: What happens if we catch a false positive? How does the user analyze that file?

GordonDiggs:
> tested locally on 8 repos

Which repos? Were any of them particularly large (lots of files, large files, etc)?

lib/batch_sanitizer.js
ABaldwinHunter: no reason. Want me to switch to `forEach` instead?

lib/batch_sanitizer.js
ABaldwinHunter: That's a valid concern. In manual testing I didn't find false positives, but the experience would be not getting a file analyzed:

- file reported as `skipped` in stderr output for engine on builds page
- no other eslint issues reported for that file on .com
- file still rated

ABaldwinHunter:
@GordonDiggs tested on a variety of sizes - shortest run took 4 seconds, longest 16 minutes (mocha).

Public repos from sample:
codeclimate/codeclimate-eslint
codeclimate/fixme
chaijs/chai
mochajs/mocha
mewo2/terrain
sachinchoolur/lightgallery.js

GordonDiggs:
@ABaldwinHunter We've used `relax/relax` as a sample large repo in the past - can you test there to be sure?

lib/batch_sanitizer.js
GordonDiggs: Meh - don't feel strongly. This is fine!

lib/batch_sanitizer.js
GordonDiggs: Ok - good to be aware of that as a caveat here in case anyone brings it up

GordonDiggs:
Code LGTM - curious how it affects relax

ABaldwinHunter:
@GordonDiggs Thanks! Relax looks good

lib/batch_sanitizer.js
wfleming: This was because I honestly forgot Node had `forEach`.

lib/batch_sanitizer.js
wfleming: Yeah, I think this will be something to think about if we do start seeing false positives. Either tweaking the heuristic or adding some "no really I mean it" config flag. I figure we'll cross that bridge when we get to it. In the things we tested while developing it, no "real" code  really naturally came anywhere close to crossing this threshold.

PR #115, opened 2016-08-11
Heuristically filter out minified files

ABaldwinHunter:
By default we filter out minified files that are identifiable as minified from
their filename (e.g. *.min.js), but many projects have minified files that
don't match those patterns, and these files can cause ESLint to segfault in a
container environment (possibly due to memory pressure).

This adds a new heuristic filter based on line length that filters out files
that appear to be minified to prevent them being analyzed. For any matching
files, a log message is sent to stderr notifying that the file is being
skipped.

@codeclimate/review @wfleming 

Change recently merged for eslint-2


ABaldwinHunter:
Related: https://github.com/codeclimate/codeclimate-eslint/pull/114/files

bin/eslint.js
```diff
@@ -10,6 +10,7 @@ console.log = console.error;
 
 var CLIEngine = require("eslint").CLIEngine;
 var docs = require("eslint").docs;
+var BatchSanitizer = require("../lib/batch_sanitizer");
```
wfleming: nit: I think this list is currently sorted.

wfleming:
Small style nit, but LGTM otherwise. I did help write this this morning, but since we already shipped this same change on `eslint-2` I don't know how much extra review you think we need here.

ABaldwinHunter:
@wfleming thanks! updated ordering.

ABaldwinHunter:
(Kept third party and our requires separate for now following pattern that was there. We can clean up further in a different PR.)

PR #116, opened 2016-08-11
Update cc badge reference

ABaldwinHunter:
Update badge url now that repo is open source on Code Climate.

@codeclimate/review


wfleming:
Good catch. :shipit:

Issue #111, opened 2016-08-01
Segmentation fault, error: (CC::Analyzer::Engine::EngineFailure) engine eslint:eslint-2 failed with status 139 and stderr


shahulhameedp:
I got this error after running `CODECLIMATE_DEBUG=1 codeclimate analyze`
Running the ESLint without codeclimate (`eslint .`) was successful in local without any issue.

I also tried with different formats and that also returning the error (`CODECLIMATE_DEBUG=1 codeclimate analyze -f json`, `CODECLIMATE_DEBUG=1 codeclimate analyze -f html`).

[DEBUG] eslint:eslint-2 engine output: 
error: (CC::Analyzer::Engine::EngineFailure) engine eslint:eslint-2 failed with status 139 and stderr

ESLint is running with the  parser.
Segmentation fault
```
[DEBUG] backtrace: /usr/src/app/lib/cc/analyzer/raising_container_listener.rb:23:in `finished'
	/usr/src/app/lib/cc/analyzer/composite_container_listener.rb:17:in `block in finished'
	/usr/src/app/lib/cc/analyzer/composite_container_listener.rb:17:in `each'
	/usr/src/app/lib/cc/analyzer/composite_container_listener.rb:17:in `finished'
	/usr/src/app/lib/cc/analyzer/container.rb:74:in `run'
	/usr/src/app/lib/cc/analyzer/engine.rb:52:in `run'
	/usr/src/app/lib/cc/analyzer/engines_runner.rb:60:in `block in run_engine'
	/usr/src/app/lib/cc/analyzer/formatters/plain_text_formatter.rb:74:in `with_spinner'
	/usr/src/app/lib/cc/analyzer/formatters/plain_text_formatter.rb:55:in `block in engine_running'
	/usr/src/app/lib/cc/analyzer/formatters/formatter.rb:18:in `engine_running'
	/usr/src/app/lib/cc/analyzer/formatters/plain_text_formatter.rb:54:in `engine_running'
	/usr/src/app/lib/cc/analyzer/engines_runner.rb:59:in `run_engine'
	/usr/src/app/lib/cc/analyzer/engines_runner.rb:23:in `block in run'
	/usr/src/app/lib/cc/analyzer/engines_runner.rb:23:in `each'
	/usr/src/app/lib/cc/analyzer/engines_runner.rb:23:in `run'
	/usr/src/app/lib/cc/cli/analyze.rb:20:in `block in run'
[...]

PR #119, opened 2016-08-15
Switch base image to avoid Segfault

GordonDiggs:
Related to #111, a Segfault occurs when we hammer STDOUT with issues.
Switching from alpine to a debian-based image seems to behave better in
this scenario.

@codeclimate/review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,12 @@
-FROM mhart/alpine-node:5.4
+FROM node:6.3.1-slim
 MAINTAINER Code Climate <hello@codeclimate.com>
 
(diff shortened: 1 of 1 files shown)


pbrisbin:
LGTM. Wouldn't be the first time we've moved off alpine for and edge-case like this.

Do you have a reliable reproduction case that is resolved with this change?

GordonDiggs:
@pbrisbin yep I reproduced this on my machine earlier and have been working with a repo locally

dblandin:
Nice. LGTM.

ABaldwinHunter:
Awesome. Nice find

PR #120, opened 2016-08-16
Update babel-eslint & eslint-flowtype

wfleming:
We're a bit behind on these. Flowtype in particular has new rules we
weren't supporting.

Looks like we hadn't actually filled in the shrinkwrap file before now,
either, so that's done.

cc @codeclimate/review


wfleming:
oops, wrong base branch.

PR #109, opened 2016-07-29
added ignorePath config option

GarthDB:
Made it possible to use `ignorePath` from the [CLIEngine](http://eslint.org/docs/1.0.0/developer-guide/nodejs-api#cliengine). Super helpful when you are already ignoring the files you don't want to lint in different file like `.gitignore`.

I also cleaned up a couple eslint errors in the project.

PR #122, opened 2016-08-17
Code cleanup

pbrisbin:
- Update eslint configuration
- Reduce complexity via temporary variable
- Consistent brace style

Extracted from #109. Will cherry-pick into eslint-2 after merging.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,10 @@
 {
   "env": {
-    "node": true
+    "node": true,
+    "mocha": true
   },
   "rules": {
+    "block-spacing": 2,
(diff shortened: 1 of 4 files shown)


.eslintrc
```diff
@@ -15,6 +16,7 @@
     "no-use-before-define": 2,
     "radix": 2,
     "semi": 2,
+    "block-spacing": 2,
```
wfleming: 💄 I think the rules here are currently sorted alphabetically?

Also should this rule be in our styleguide?

wfleming:
Small style nit, LGTM otherwise.

PR #123, opened 2016-08-17
Support for ignore_path in engine config

pbrisbin:
Makes it possible to use ignorePath from the CLIEngine. Useful when you
are already ignoring the files you don't want to lint in different file
(.e.g .gitignore).

/cc @codeclimate/review

Extracted from #109.


GordonDiggs:
LGTM

PR #128, opened 2016-09-07
channel/eslint 2

ABaldwinHunter:


ABaldwinHunter:
@codeclimate/review

pbrisbin:
If this is the new the deployment branch, with no other changes, I don't see any reason to review. Ship away!

circle.yml
```diff
@@ -20,7 +20,7 @@ test:
 
 deployment:
   registry:
-    branch: master
+    branch: channel/eslint-2
```
pbrisbin: (And of course now I see something to note :))

I don't think we want to have to edit the circle.yml in all the deployment branches. I think this value can be something like `master|channel/*` to trigger deployment for all the branches we want it to.

circle.yml
```diff
@@ -20,7 +20,7 @@ test:
 
 deployment:
   registry:
-    branch: master
+    branch: /master|channel\/*/
```
dblandin: I think you're missing a character specifier within the regex here:
```
/master|channel\/[\w-]+/
```

dblandin:
Closed this PR but the branch will live on as it now maps to an active engine channel.

Going forward, channels will not have persistent PRs on their own. Code underlying specific channels may be viewed via the corresponding branch. Future PRs should be based off of the branch corresponding to the appropriate channel.

For example, a PR suggesting a change to the `eslint-3` channel should have a base branch of `channel/eslint-3`.

PR #148, opened 2016-11-03
Revert "Add eslint-plugin-import resolvers for node and webpack"

dblandin:
This reverts commit ac35f7ea4984c3124ad8451d952cb3c41a75c171.

Unfortunately, this behavior doesn't seem to work when modules are
imported in source code but not installed via npm, as is the case when
running static analysis builds on codeclimate.com.

2016-11-03-172051_999x478_scrot
2016-11-03-172034_1081x284_scrot

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -29,16 +29,6 @@
       "from": "ajv-keywords@>=1.0.0 <2.0.0",
       "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-1.1.1.tgz"
     },
-    "align-text": {
-      "version": "0.1.4",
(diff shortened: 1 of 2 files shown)


dblandin:
Yep, in particular the `import/no-unresolved` check.

I think some of these rules could be re-introduced along with a new blacklist that prevents actual module lookup.

AndrewRayCode:
We're currently hitting this problem with Webpack, eslint and codeclimate. Can you add more context to this? I'm confused about this revert. Requiring packages that aren't installed via npm should clearly raise a linting error. In the case of the screenshot, glob and stylelint aren't default node packages, so if their source isn't explicitly stated, why wouldn't you want to error?

PR #151, opened 2016-11-09
Upgrade ESLint & some packages

### LINKED ISSUE #150: Syntax validation
Hello,

I'm getting this error while running in the .com engine.
```
Configuration for rule "comma-dangle" is invalid:
	Value "[object Object]" must be an enum value.
```

My `comma-dangle` rule:
```
"comma-dangle": ["warn", {
  "arrays": "always-multiline",
  "objects": "always-multiline",
  "functions": "only-multiline"
}],
```

Probably a ESLint version issue? I couldn't find which version the engine was running on.
Upgrade time? :)

wfleming:
* We're several point releases behind on ESLint
* After updating just that, I got some errors from shrinkwrap for two
  other packages we support, so I upgraded those too.

This should fix #150 

cc @codeclimate/review


dblandin:
LGTM

PR #153, opened 2016-11-23
Update package.json

leftees:
Add support for eslint-plugin-react-intl plugin to eslint-2 channel.

PR #154, opened 2016-11-23
Update package.json

leftees:
Add support for eslint-plugin-react-intl plugin to eslint-3 channel.


maxjacobson:
Same missing step as https://github.com/codeclimate/codeclimate-eslint/pull/153#issuecomment-262617274

Not sure if this is documented anywhere, but probably should be in the README of this repo

PR #157, opened 2016-11-29
Fix .codeclimate.yml example in contributing

SebastienElet:
It come from the documentation : https://docs.codeclimate.com/docs/example-codeclimateyml


wfleming:
Thanks for the correction @SebastienElet!

PR #159, opened 2016-12-13
Switch to yarn for installing dependencies

maxjacobson:
This seems more intuitive to me today, and I hope will make it easier for contributors who want to add or upgrade plugins.

If we agree this is a good approach to take, I'll update the other channels as well.


wfleming:
I am extremely happy to see this! I've been wanting to make time to do this here & in some other node-based repos for a while.

Looks like something odd is happening in the docker build on CI with a yarn dep?

CONTRIBUTING.md
```diff
@@ -1,6 +1,24 @@
 # Contributing
 
+## Adding plugins
+
+When adding or upgrading a plugin, we should see changes in both `package.json`
+and `yarn.lock`.
+
+You'll need to [install yarn](https://yarnpkg.com/en/docs/install).
```
maxjacobson: Ah, right. I was just removing the `shrinkwrap` task from the Makefile, but I should be able to adapt it instead, right?

maxjacobson: oh, I might see my mistake -- is it running those 3 docker commands in 3 separate containers or something like that? 😞 

I did it on multiple lines because there are multiple files we expect to change as part of these commands...

maxjacobson:
@wfleming yep, got it working by using nightly yarn. They've fixed a bug, but it's only available in https://github.com/yarnpkg/yarn/releases/tag/v0.18.0 which is a pre-release (PR 1837)

CONTRIBUTING.md
wfleming: > For me, running yarn outside of the container is a better experience. What are the reasons you'd prefer not to?

For the same reason I don't run `bundle` locally to update `Gemfile.lock`, and for the same reasons we have images for some repos that aren't even deployed in containers: I don't need to even install `node` (or `ruby`, or whatever) locally, and operations on the repo should pretty much only ever depend on `docker`.

> oh, I might see my mistake -- is it running those 3 docker commands in 3 separate containers or something like that? 😞

Yep. 

Since you want to modify more than one file, I think what you really want to do is to mount in `$PWD` to `/usr/src/app`.

You could also just make a `bin/yarn` stub in this repo to run yarn with given args with the correct mounting, rather than making a make task for every possible command. We do this in several other repos for related tools.

maxjacobson:
@wfleming cool, that all made sense. Updated

bin/yarn
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+
```
maxjacobson: @wfleming question

I was tempted to add `make image` here. The reason: maybe you just cloned the repo and want to add a plugin. If you haven't run `make image`, it will pull the image from dockerhub and then try to run the command. However, we have multiple channels, and it will pull `latest` rather than `eslint-3`, and latest doesn't have yarn yet, and is sort of a different thing.

`make image` gets around that problem, but has the downside that it will need to re-build the image each time the `package.json` or `yarn.lock` change, so it's slow to run it on subsequent yarn operations.

Everything does work nicely as long as you independently run `make image` yourself before you start using `bin/yarn`... so maybe I should just say to do that in `CONTRIBUTING.md`?

wfleming: I think `make image` should always be step 1 for local dev, so that's a good thing to call out in doc. I don't think a bin stub like this should make the image itself (we don't do that in other similar bin stubs).

The thing about channels & local images & such is a shame, but that's messy *today* as well, so I think we can leave that yak hairy for the time being.

PR #160, opened 2016-12-14
Switch to yarn for installing dependencies (eslint-2)

maxjacobson:
Sibling PR to https://github.com/codeclimate/codeclimate-eslint/pull/159


maxjacobson:
@wfleming btw - I'm having a harder time adding yarn to master branch because its base image is alpine linux, and there isn't a package set up in their package manager. I also tried (for ~10 minutes) to use [their install script](https://yarnpkg.com/en/docs/nightly) in the Dockerfile but it didn't feel like it wanted to work. Thoughts on how to proceed?

maxjacobson:
OK. I'm inclined to leave it as-is for now, then.

PR #168, opened 2017-01-23
Channel/eslint 3 create react app

danielchangNYC:
Adds dependencies required for `create-react-app` eslint configs.

Addresses #168


maxjacobson:
@danielchangNYC hey, thanks for this PR. Is the goal that you can eventually use [eslint-config-react-app](https://github.com/facebookincubator/create-react-app/tree/master/packages/eslint-config-react-app)? I think we may also need to add that as an explicit dependency if so, unless you're planning to pull in a config file via [external configuration](https://codeclimate.com/changelog/582495c32c33066f1b00191d). WDYT?

package.json
```diff
@@ -8,14 +8,15 @@
...
-    "babel-eslint": "^6.1.2",
+    "babel-eslint": "^7.0.0",
     "eslint": "3.11.1",
     "eslint-config-airbnb": "^11.0.0",
     "eslint-config-airbnb-base": "^7.0.0",
     "eslint-config-angular": "^0.5.0",
     "eslint-config-ember": "^0.3.0",
     "eslint-config-google": "^0.7.1",
     "eslint-config-hapi": "^10.0.0",
+    "eslint-config-react-app": "^0.5.0",
```
danielchangNYC: @maxjacobson is this what you mean by adding an explicit dependency? Or does this dependency need to be defined elsewhere too?

maxjacobson: heh, whoops, I missed that 😅

maxjacobson:
OK this LGTM, will plan to ship this shortly

maxjacobson:
This is live now :)

danielchangNYC:
Thanks @maxjacobson !

PR #170, opened 2017-01-27
added eslint-plugin-xogroup as part of codeclimate eslint engine acce…

lamchakchan:
**Context**

As a customer of Code Climate, we would like to extend the `codeclimiate-eslint` engine to accept our custom `eslint-plugin-xogroup` plugin.  We have many JS projects analyzed under the Code Climate toolset.  It is at the point where we need a single source of truth for our rules definition which `eslint` supports.  Since Code Climate does not extensibly allow for custom configs for plugins with `eslint` support, our Code Climate rep informed us we were free to make a PR against this engine to get our plugin added.  Our `eslint-plugin-xogroup` plugin has been NSP scanned.  Please accept that as security compliant.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "eslint-plugin-promise": "1.3.1",
     "eslint-plugin-react": "3.16.1",
     "eslint-plugin-standard": "1.3.1",
+    "eslint-plugin-xogroup": "^1.0.5",
     "glob": "5.0.14"
   },
   "devDependencies": {


lamchakchan:
I'm not positive that all was required was to touch up the `package.json` file.  If there is another file that needs to be updated in order to accept our plugin, please ping back.

maxjacobson:
Hi @lamchakchan, thanks for the PR!

> If there is another file that needs to be updated in order to accept our plugin, please ping back.

We recommend using the eslint-3 channel of the engine. If you'd like to update this PR to target the `channel/eslint-3` branch, you can follow [the CONTRIBUTING.md](https://github.com/codeclimate/codeclimate-eslint/blob/channel/eslint-3/CONTRIBUTING.md) file on that branch.

PR #174, opened 2017-02-03
Upgrade eslint

maxjacobson:
```
make
bin/yarn upgrade eslint
```

We'd like to pull in a fix that prevents using this rule:
http://eslint.org/docs/2.0.0/rules/generator-star-spacing

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "babel-eslint": "^7.0.0",
-    "eslint": "3.11.1",
+    "eslint": "^3.14.1",
     "eslint-config-airbnb": "^11.0.0",
     "eslint-config-airbnb-base": "^7.0.0",
     "eslint-config-angular": "^0.5.0",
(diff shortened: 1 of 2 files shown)


wfleming:
I'm confused that this is for ESLint 3 but your docs link has `2.0.0` in the URL -- is that URL version not related to the ESLint version or something? I also don't see that rule mentioned in recent entries in ESLint's [changelog](https://github.com/eslint/eslint/blob/master/CHANGELOG.md).

The upgrade seems fine in general, so carry on, just curious about the motivating change.

maxjacobson:
Ah, sure. The error we're seeing is the same one reported in this issue: https://github.com/eslint/eslint/issues/6274

The [final comment](https://github.com/eslint/eslint/issues/6274#issuecomment-266782379) as of this writing suggests they think it's fixed in the latest version of eslint. I don't have total context on what the fix was, but I think it was something more general, and not specifically for this rule, but which also fixed this rule.

wfleming:
Cool cool, thanks 👍

wfleming:
Small nit of your commit message: the relevant fix doesn't prevent using the rule, it allows use of the rule. The *bug* prevented using the rule in some circumstances. (If I understood all of this correctly.)

maxjacobson:
Good catch :). I updated to clarify what I was trying to say originally

PR #175, opened 2017-02-14
Add support for eslint prettier plugin

leftees:
@maxjacobson Can you let me know when this is on production? Want to make sure I remember to update the docs.


maxjacobson:
@leftees this is live now!

PR #177, opened 2017-02-17
Config upgrader

pointlessone:
Automatically upgrade pre-ESLint 3 config to ESLint 3-compatible format.

Upgrade instructions are directed to stderr.

image

### DIFF
--- a/lib/config_upgrader.js
+++ b/lib/config_upgrader.js
@@ -1,0 +1,272 @@
+"use strict";
+
+const Config = require("eslint/lib/config")
+  , merge = require("eslint/lib/config/config-ops").merge
+  , path = require("path")
+  , stringify = require("json-stable-stringify")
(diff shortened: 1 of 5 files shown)


lib/config_upgrader.js
```diff
@@ -0,0 +1,272 @@
+"use strict";
```
GordonDiggs: I'm a bit confused by this - I thought that ESLint had built-in functionality we could use for this. Why do we need this file?

pointlessone: As far as I can tell there's no code for config upgrades in the ESLint repo or in any other repo under ESLint org. Unfortunately.

There are a few upgrade guides for each major version of ESLint. They suggest manual config upgrades. This file performs what those guides recommend doing in-memory. This code also gives specific instructions for the end-user what they need to do to their configs to bring them up to the latest version.

PR #178, opened 2017-02-23
ESLint 3 preparation for master

pointlessone:
This is a preparation of `channel/eslint-3` branch for merge into master. The HEAD of this branch is effectively the same as `channel/eslint-3`.

This branch contains all commits from ESLint 3 branch except for merge commits and shared commits. Some commits are a bit different to the original ones because of conflict resolution during cherry-picking.

Because I cherry-picked commits git/GH shows me as the committer on all these commits. It's possible to amend commits so that their committers were the same as authors. Let me know if it's important.


pointlessone:
I'm not sure what to do with CC failure as this PR effectively has no new code. It just moves it around a bit.

maxjacobson:
That sounds like a good opportunity to use the PR approval feature :)

GordonDiggs:
This LGTM - we should wait to merge it until @maxjacobson resolves the upgrader issue

lib/eslint-patch.js
```diff
@@ -0,0 +1,43 @@
...
+var meld = require('meld');
+var docs = require('./docs');
+var Config = require("eslint/lib/config");
+var ConfigUpgrader = require('./config_upgrader');
+
+var supportedPlugins = ['react', 'babel'];
+
+module.exports = function patcher(eslint) {
+
+  meld.around(eslint.CLIEngine, 'loadPlugins', function(joinPoint) {
```
filipesperandio: @GordonDiggs @pointlessone Hey guys! I am trying to understand the reasons for this. Digging at the `eslint` source, it seems the `loadPlugings` function was removed on version 2.0, we are currently supporting version  3.14.1... is this a leftover from some experimentation?

GordonDiggs: This was likely added when we supported ESLint 2 on the way to supporting ESLint 3. Is there a similar function in ESLint 3 that we can meld?

filipesperandio: Looking at the history it seems to be added while upgrading to 3, that's why I am confuse.
Anyways I am digging into eslint to understand where did this logic go.

PR #182, opened 2017-02-24
Fix upgrader with specified ESLint config

pointlessone:
Upgrader improperly kept detecting ESLint config file when engine had a specific config file specified.

This behaviour resulted in providing upgrade instructions for the wrong config file or in worse case trowing an exception of missing config file.


maxjacobson:
Awesome job! This looks like it resolves the upgrade output and also prevents attempting to use other config files when the user has explicitly provided a path to an alternate config file in their `.codeclimate.yml`

PR #195, opened 2017-03-06
Add eslint-plugin-vue and eslint-plugin-html

JiriChara:
Hey,

I tried to build my repository on codeclimate and it seems like that `eslint-plugin-vue` and `eslint-plugin-html` is missing.

I hope this PR will fix the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "eslint-plugin-ember-suave": "^1.0.0",
     "eslint-plugin-flowtype": "^2.21.0",
     "eslint-plugin-hapi": "^4.0.0",
+    "eslint-plugin-html": "^2.0.1",
     "eslint-plugin-immutable": "^1.0.0",
     "eslint-plugin-import": "^2.0.1",
     "eslint-plugin-import-order": "^2.1.4",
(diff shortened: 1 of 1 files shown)


dblandin:
Hey @JiriChara ,

Thanks for the PR! Are there any category mappings to specify for these plugins?

https://github.com/codeclimate/codeclimate-eslint/blob/master/lib/check_category_mapping.json

Looks like this PR might also need a rebase.

JiriChara:
Hey @dblandin,

I have rebased to latest greatest master, but  I am not sure if any category mappings are required. How can I find it out?

dblandin:
If the plugins only emits `Style` issues, which is the default, then you should be all set. If the plugins check for issues of complexity, clarity, security, etc, then some mappings would be beneficial.

Here's a list of the categories that the we currently support: https://github.com/codeclimate/spec/blob/master/SPEC.md#categories

JiriChara:
@dblandin both plugins are checking styling only right now. I think they would like to add check for unclosed tags, but it's not yet implemented.

JiriChara:
@dblandin Do you think we can get this merged?

dblandin:
@JiriChara FYI: I just released an update including the plugins from this PR. Should be available via the CLI and codeclimate.com now. Let me know if you run into any issues!

JiriChara:
@dblandin works like a charm. Thanks!

PR #198, opened 2017-03-06
Avoid errors when issue messages not how we expect

maxjacobson:


dblandin:
Before skipping on the complexity fingerprint implementation, I think it would be best to understand the error case better. How did the error message change that caused this to break?

maxjacobson:
Fair enough -

The greenkeeper update commit (https://github.com/codeclimate/codeclimate-eslint/commit/4a5283411015387d8a9af009824234902ac285e0) incidentally upgraded eslint from 3.14.1 to 3.17.0: https://github.com/eslint/eslint/compare/v3.14.1...v3.17.0

In [an affected OSS repo](https://github.com/chrishulton/whackacell), I tracked it down to this difference:

[before](https://github.com/eslint/eslint/blob/v3.14.1/lib/rules/complexity.js#L94) and [after](https://github.com/eslint/eslint/blob/v3.17.0/lib/rules/complexity.js#L98)

maxjacobson:
It looks like we wanted custom fingerprinting logic specifically for this rule, so if we want to upgrade to a newer eslint we'll need to find a new strategy. Agreed this change isn't appropriate, after checking that.

maxjacobson:
But, what to do?

PR #202, opened 2017-03-09
Easier debugging

filipesperandio:
`make test NPM_TEST_TARGET=test.debug`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,17 +2,18 @@
 
 IMAGE_NAME ?= codeclimate/codeclimate-eslint
 
-run_interactively := docker run -ti --volume $(PWD):/usr/src/app --rm $(IMAGE_NAME) sh -c
+NPM_TEST_TARGET ?= test
 
 image:
 	docker build --rm -t $(IMAGE_NAME) .
 
-test:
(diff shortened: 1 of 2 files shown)


filipesperandio:
@pbrisbin What your taste says?

pbrisbin:
I might do something like this:

```make
NPM_TEST_TARGET ?= test

test:
    docker run --rm \
      --volume ... \
      --workdir ... \
      $(IMAGE_NAME) npm $(NPM_TEST_TARGET)
```

Using `--workdir` (something we didn't realize existed when this was first written) makes things a bit easier.

```console
make test NPM_TEST_TARGET=test.debug
```

filipesperandio:
Interesting. It keeps the Makefile clean.
I was going to say that the WORKDIR is already to what we expect here, then I realize we actually have it duplicated [here](https://github.com/codeclimate/codeclimate-eslint/blob/master/Dockerfile#L34)

pbrisbin:
Yeah, for some reason the SPEC says `WORKDIR` must be `/code`, so we set it to that at the end. We have the one at the top to make the commands in between easier.

PR #203, opened 2017-03-09
Improve layer caching for `docker build`

filipesperandio:


Dockerfile
```diff
@@ -1,15 +1,17 @@
...
-WORKDIR /usr/src/app
-COPY package.json yarn.lock /usr/src/app/
-
 RUN apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg && \
     echo "deb http://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
     apt-get update && \
-    apt-get install -y git jq yarn && \
-    yarn install && \
-    git clone https://github.com/eslint/eslint.git && \
+    apt-get install -y git jq yarn
```
GordonDiggs: for size, can you install git and jq below right before the `git clone`?

filipesperandio: :+1:

PR #208, opened 2017-03-15
Upgrade ESLint dependency to 3.17.1

dblandin:
- Upgrade ESLint, plugins, and shared configs
- Update complexity issue fingerprint to account for anonymous functions
- Update configuration file validation


lib/compute_fingerprint.js
```diff
@@ -23,9 +23,13 @@ Fingerprint.prototype.compute = function() {
...
 Fingerprint.prototype.strippedMessage = function(message) {
   var regex = /Function '\S+'/
-    , stripped = message.match(regex);
+    , match = message.match(regex);
 
-  return stripped[0];
+  if (match) {
+    return match[0];
+  } else {
+    return "Anonymous Function";
```
maxjacobson: How will this behave when there are multiple, complex anonymous functions in one file?

maxjacobson: Ah, I see you acknowledged that in your commit message. I suppose it's the best we can do

lib/validate_config.js
```diff
@@ -3,7 +3,7 @@ var CLIEngine = require("eslint").CLIEngine
 
 module.exports = function(configPath) {
   if (configPath) {
-    return true;
+    fs.existsSync(configPath);
```
maxjacobson: doesn't this need a `return` at the start?

Makefile
```diff
@@ -16,4 +16,5 @@ test: image
 citest:
 	docker run --rm \
 		--workdir /usr/src/app \
+		--volume /tmp:/tmp \
```
maxjacobson: this seems fine, but I'm curious what motivated this. (I always appreciate more context in commit messages for this kind of thing)

lib/validate_config.js
dblandin: Hmmm, tests pass but a return seems wise.

Makefile
dblandin: Build wasn't happy without it:

```
  1) validateConfig returns false if no files exist:
     Uncaught AssertionError: expected undefined to equal true
      at test/validate_config_test.js:15:44
      at node_modules/temp/lib/temp.js:252:7
      at FSReqWrap.oncomplete (fs.js:123:15)

  2) validateConfig Uncaught error outside test suite:
     Uncaught TypeError: Cannot read property 'currentRetry' of undefined
```

https://circleci.com/gh/codeclimate/codeclimate-eslint/476

Happy to add a note in the commit description. The `temp` module seemed happier with the volume mount.

PR #210, opened 2017-03-15
Remove file is ignored by `.eslintignore` warnings

dblandin:
These warnings generate a lot of noise and obfuscate useful error
messages in errored build output.

We can let ESLint handle ignored file behavior instead of pruning the
paths up front ourselves.

### DIFF
--- a/bin/eslint.js
+++ b/bin/eslint.js
@@ -97,12 +97,7 @@ function isFileWithMatchingExtension(file, extensions) {
 }
 
 function isFileIgnoredByLibrary(file) {
-  var ignored = cli.isPathIgnored(file);
-  if (ignored) {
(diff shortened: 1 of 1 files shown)


dblandin:
> This actually removes the warning and the check, but we want to keep the check, just lose the messaging, right?

Maybe? Is the concern that we overload the number of paths we pass to ESLint? Can we just let ESLint handle the ignored file logic instead of intercepting that up front and pruning the paths within the engine?

pbrisbin:
I don't know the history or the answer to those questions, unfortunately. I would suggestion you do one of:

1. Be safe and just remove the warning
1. Remove the warning here and open another PR to remove the check and solicit more discussion there
1. Pause on this PR as-is until you can get confidence it's OK to merge

PR #212, opened 2017-03-15
Integration tests setup

filipesperandio:
Is this valuable?


lib/eslint.js
```diff
@@ -0,0 +1,259 @@
...
+  var configPath = runOptions.configPath || "/config.json";
+  var codeDir = runOptions.dir || "/code";
+
+  var cli; // instantiation delayed until after options are (potentially) modified
+  var debug = false;
+  var ignoreWarnings = false;
+  var ESLINT_WARNING_SEVERITY = 1;
+
+  var output = {
+    stdout: [],
```
nporteschaikin: Why not send the `process` object (or, maybe `process.stdout`, `process.stderr`, and `process.exit`) as constructor arguments or as options? Then it'd be easy to stub out dummies for each.

lib/eslint.js
```diff
@@ -0,0 +1,259 @@
...
+  };
+
+  // a wrapper for emitting perf timing
+  function runWithTiming(name, fn) {
+    const start = new Date();
+    const result = fn();
+
+    if (debug) {
+      const duration = (new Date() - start) / 1000;
+      console.error("eslint.timing." + name + ": " + duration + "s");
```
nporteschaikin: Wouldn't we want to send this to our `stderr` array, for the sake of consistency?

lib/eslint.js
filipesperandio: Good point, that was the first thing I thought of to allow for testing. Will try out some of the suggestions.

lib/eslint.js
filipesperandio: Probably... wasn't changing much (trying not to) in the move from `bin/`. There are actually lots of other places calling `console.error` as well.

filipesperandio:
@nporteschaikin I would prefer to keep it as a single PR once there is no unit test for the class I am refactoring, the integration tests at least make me a bit more confident I am not breaking anything.
On top of that, most of the changes were made to allow for testability.

Anyways, made some adjustments. Appreciate if you can give another shot.

lib/eslint.js
filipesperandio: I went the other way and ditch all `process.stderr` in favor of `console.error`.

lib/eslint.js
```diff
@@ -0,0 +1,243 @@
...
+  var ESLINT_WARNING_SEVERITY = 1;
+
+  // a wrapper for emitting perf timing
+  function runWithTiming(name, fn) {
+    const start = new Date();
+    const result = fn();
+
+    if (debug) {
+      const duration = (new Date() - start) / 1000;
+      console.error("eslint.timing." + name + ": " + duration + "s");
```
nporteschaikin: This still feels fragmented to me. I'd go one of two ways:

- Pass `stdout` and `stderr` as constructor arguments, or
- Use `stdout` and `stderr` directly in the function and replace each in integration tests.

WDYT?

Issue #214, opened 2017-03-16
Odd warnings due unsupported rules when extending a config


lucasmazza:
When extending the `airbnb` config, none of the `eslint-plugin-import` rules are found:

```
$  codeclimate  analyze
Starting analysis
Running eslint: Done!

== index.js (17 issues) ==
1: Definition for rule 'import/no-duplicates' was not found [eslint]
1: Definition for rule 'import/no-unresolved' was not found [eslint]
1: Definition for rule 'import/no-named-as-default' was not found [eslint]
1: Definition for rule 'import/no-named-as-default-member' was not found [eslint]
1: Definition for rule 'import/no-extraneous-dependencies' was not found [eslint]
1: Definition for rule 'import/no-mutable-exports' was not found [eslint]
1: Definition for rule 'import/no-amd' was not found [eslint]
1: Definition for rule 'import/first' was not found [eslint]
1: Definition for rule 'import/no-named-default' was not found [eslint]
1: Definition for rule 'import/extensions' was not found [eslint]
1: Definition for rule 'import/newline-after-import' was not found [eslint]
1: Definition for rule 'import/prefer-default-export' was not found [eslint]
1: Definition for rule 'import/no-absolute-path' was not found [eslint]
1: Definition for rule 'import/no-dynamic-require' was not found [eslint]
1: Definition for rule 'import/no-webpack-loader-syntax' was not found [eslint]
1: Definition for rule 'import/export' was not found [eslint]
1: 'React' is defined but never used. [eslint]
```
Adding `plugins: ['import']` to my config raises another warnings for different rules from the airbnb config. Locally this doesn't happen when the image is rolledback to the **b457** tag (the previous commit from the #204 merge).

[Sample repository here.](https://github.com/lucasmazza/eslint-extends-test)

ahoym:
I'm actually seeing something similar as well. Here's a screenshot I took while using the codeclimate chrome plugin:

<img width="710" alt="screen shot 2017-03-16 at 2 28 55 pm" src="https://cloud.githubusercontent.com/assets/5249615/24019668/261a7718-0a56-11e7-957f-bdf458929ae6.png">

Which pops up in almost every file that adds or modifies an `import`

dblandin:
@lucasmazza @ahoym We're on it. Thank you for the detailed issue and example repo!

Looks like #204 may be interfering with our vendored plugins. Working on a fix now.

dblandin:
[...]

PR #215, opened 2017-03-16
Does not raise error for invalid rules - Fix #213

### LINKED ISSUE #192: Don't error if we encounter a plugin we don't support
We have a [list of supported ESLint plugins](https://github.com/codeclimate/codeclimate-eslint/blob/master/package.json#L11), but there are many we don't support. Currently, having a plugin in your config that is not supported will result in an error. We should instead print a warning to stderr that says that we are skipping it

### LINKED ISSUE #213: Don't Error when no rules are configured
If there is a blank config file, or if the config file _only_ inherits from an external configuration, the ESLint engine can error with the following error:

```
No rules are configured. Make sure you have added a config file with rules enabled.
```

This should be printed to STDERR and not error the build

filipesperandio:
Bring back functionality reverted on #216.
Fix #192, #213 and #214.

The main intention is to not raise errors about unsupported plugins/rules/modules and let ESLint proceed without them.

All known scenarios with problematic configurations were added as integration tests.


dblandin:
I observed the same issues described in #214 locally after building an image from this branch. I've since reverted #204 in #216, so hopefully this PR can fix the issue while reintroducing those desired updates.

filipesperandio:
@dblandin I might not have the big picture here, but I was expecting we could just rollback the `builder` instead.  https://github.com/codeclimate/builder/pull/893/files

maxjacobson:
@filipesperandio we try to follow the principle that whatever is on master is safe to deploy to production to prevent accidents. Often we revert things and then revert the revert and add a fix.

dblandin:
> I might not have the big picture here, but I was expecting we could just rollback the builder instead.

We typically prefer to revert the specific engine change and roll forward for a few reasons:

- Anything on the `master` branch should be "ready to deploy"
- Other PRs unrelated to the problematic commits can be merged and deployed in the meantime
- Passing CI builds on `master` trigger an image push to Docker Hub, which the CLI pulls from directly.

CLI users get whatever is latest on Docker Hub. If we don't revert the engine-specific PR, the engine for CLI users would remain affected. If we only rolled back the builder version used in production, we would lose any other engine updates made after we introduced the problematic change. If we only reverted the specific builder PR bumping the eslint engine image version, we would lose any following unrelated changes made to the eslint engine specifically.

Hope that makes sense!

dblandin:
@filipesperandio Hey! Sorry, I think I misunderstood the intent of this PR based on our conversation in Slack yesterday. This PR does seem to fix #213 but #204 should be fixed and re-introduced before moving forward on this PR.

filipesperandio:
@dblandin @maxjacobson Makes perfect sense.

@dblandin Getting back to this, now... I will make sure to address all those problems.

filipesperandio:
@dblandin I think I misunderstood the issue #214 too.

dblandin:
@filipesperandio Is this ready for review?

filipesperandio:
@dblandin It is now.

dblandin:
@filipesperandio This looks good to me! 👍 Looks like it needs a rebase.

Before merging, could you update the PR description to briefly explain the new changes introduced in this PR following the earlier reverts?

filipesperandio:
@dblandin Sure thing.

Issue #173, opened 2017-02-03
Add support for eslint-plugin-html


artn:
I am using RiotJS with ES6 and using [eslint-plugin-html](https://github.com/BenoitZugmeyer/eslint-plugin-html) for linting.

When build the code on codeclimate, error was raised.

```
File `.eslintignore` ignored because of your .eslintignore file.
File `.eslintrc.js` ignored because of your .eslintignore file.
File `.eslintrc` ignored because of your .eslintignore file.
File `.gitignore` ignored because of your .eslintignore file.
File `.codeclimate.yml` ignored because of your .eslintignore file.
/usr/src/app/node_modules/eslint/lib/config/plugins.js:137
                throw err;
                ^

Error: Failed to load plugin html: Cannot find module 'eslint-plugin-html'
    at Function.Module._resolveFilename (module.js:455:15)
    at Function.Module._load (module.js:403:25)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
    at Object.load (/usr/src/app/node_modules/eslint/lib/config/plugins.js:129:26)
    at Array.forEach (native)
    at Object.loadAll (/usr/src/app/node_modules/eslint/lib/config/plugins.js:151:21)
    at Object.load (/usr/src/app/node_modules/eslint/lib/config/config-file.js:504:21)
    at loadConfig (/usr/src/app/node_modules/eslint/lib/config.js:63:33)
    at new Config (/usr/src/app/node_modules/eslint/lib/config.js:226:38)

```

Could you add support for this please?

dblandin:
We'd be happy to look at a pull request!

https://github.com/codeclimate/codeclimate-eslint/blob/master/CONTRIBUTING.md

artn:
When I executed `bin/yarn add eslint-plugin-html` below error was raised.

```
$ bin/yarn add eslint-plugin-html
yarn add v0.21.3
[1/4] Resolving packages...
[2/4] Fetching packages...
error Could not write file "/usr/src/app/yarn-error.log": "EACCES: permission denied, open '/usr/src/app/yarn-error.log'"
error An unexpected error occurred: "EACCES: permission denied, unlink '/usr/src/app/node_modules/.yarn-integrity'".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

So I executed below command and found that in my docker environment(MacOS host and  using docker-machine with VirtualBox driver), login user is **app** but mounted volume's owner is **node**.

```
[...]

PR #217, opened 2017-03-16
Update yarn executable

dblandin:
- Update bin/yarn executable
  - Specify current user id/group within container to fix write permission errors
  - Simplify docker run invocation
- Add shellcheck engine to codeclimate analysis

Addresses https://github.com/codeclimate/codeclimate-eslint/issues/173#issuecomment-286965537

```
$ bin/yarn add eslint-plugin-html
yarn add v0.21.3
[1/4] Resolving packages...
[2/4] Fetching packages...
error Could not write file "/usr/src/app/yarn-error.log": "EACCES: permission denied, open '/usr/src/app/yarn-error.log'"
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
error An unexpected error occurred: "EACCES: permission denied, unlink '/usr/src/app/node_modules/.yarn-integrity'".
```

### DIFF
--- a/bin/yarn
+++ b/bin/yarn
@@ -2,5 +2,14 @@
 
 set -e
 
-IMAGE_NAME=${IMAGE_NAME:-codeclimate/codeclimate-eslint}
-docker run --rm --volume "$PWD:/usr/src/app" "$IMAGE_NAME" sh -c "cd /usr/src/app && yarn $*"
+case "$(uname -s)" in
+ "Darwin") user="root:root" ;;
+ *) user="$(id -u):$(id -g)" ;;
+esac
+
+docker run \
+  --rm \
+  --user "$user" \
+  --volume "$PWD:/usr/src/app" \
+  --workdir "/usr/src/app" \
+  codeclimate/codeclimate-eslint yarn "$@"
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -6,10 +6,12 @@ engines:
     config:
       languages:
       - javascript
+  shellcheck:
+    enabled: true
 ratings:
   paths:
-  - bin/eslint.js
-  - lib/*.js
+  - bin/
+  - lib/**/*.js
 exclude_paths:
 - "node_modules/**"
 - "test/**"


bin/yarn
```diff
@@ -2,5 +2,9 @@
 
 set -e
 
-IMAGE_NAME=${IMAGE_NAME:-codeclimate/codeclimate-eslint}
-docker run --rm --volume "$PWD:/usr/src/app" "$IMAGE_NAME" sh -c "cd /usr/src/app && yarn $*"
+docker run \
+  --rm \
+  --user "$(id -u):$(id -g)" \
```
wfleming: This may not work as desired for Mac users: I know the Docker for Mac VM runs as root, and I think I ran into some issues with user perms and mounts working with that at one point. 

Come to think of it, that might have been for stuff within a Dockerfile, so this might work fine. Might want to have a Mac user confirm before merging, though.

wfleming: I think I confirmed locally this breaks with Docker for Mac. Got this running on this branch:

```
[codeclimate-eslint]  bin/yarn upgrade                                                         devon/update-yarn-executable☃ ✱
yarn upgrade v0.21.3
error An unexpected error occurred: "EACCES: permission denied, mkdir '/.config'".
info If you think this is a bug, please open a bug report with the information provided in "/usr/src/app/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/upgrade for documentation about this command.
```

Worked as expected on `master`.

I did some conditional fanciness in one of our repos way back to set this conditionally by platform. If you think it's worth it, you could do the same here by grepping `uname -a` for `Darwin` and using root:root on Darwin, and what you already have in other cases.

maxjacobson: I think making this cross-compatible is ideal

dblandin: Just pushed up a commit with a conditional.

@maxjacobson Would you mind trying it out?

dblandin:
Verified the mac behavior (at least via Docker for Mac) with @ABaldwinHunter. Going to go ahead and merge this on green.

PR #218, opened 2017-03-17
Improve docker layer caching and build times

dblandin:
Currently, the apt-get package install layer is invalidated anytime we
update npm dependencies. This commit splits package installation and npm
dependency installation into separate steps, which should improve docker
layer caching and speed up build times.

This commit also extracts the inline docs installation into a separate
bin script.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,32 +1,26 @@
 FROM node:6.10.0-slim
-MAINTAINER Code Climate <hello@codeclimate.com>
+LABEL maintainer "Code Climate <hello@codeclimate.com>"
 
 RUN apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg && \
     echo "deb http://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
-    apt-get update && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -4,27 +4,18 @@ MAINTAINER Code Climate <hello@codeclimate.com>
...
 
 WORKDIR /usr/src/app
-COPY package.json yarn.lock /usr/src/app/
+COPY bin/docs ./bin/docs
+COPY package.json yarn.lock ./
 
-RUN yarn install
+RUN yarn install && \
+    bin/docs && \
+    apt-get purge --yes git jq yarn && \
```
wfleming: Purging these packages in a separate layer from where they're installed is sort of useless: they're still in the earlier layers, so they still need to pulled.

If we can tolerate the size of these packages, we should just skip the `purge`. If they have a significant impact on overall size, then in order to really keep size down we'd need to do the `apt-get install` in the same layer as the other commands here, which seems contrary to your goal.

dblandin: Yep, good point. Looking into a better way of doing this...

dblandin:
@wfleming This is ready for another look when you have a chance.

PR #226, opened 2017-03-21
Use bin script to update yarn.lock via CircleCI

dblandin:
This commit updates our CI setup to use local bin script instead of the
[npm module I found][module] to update the yarn.lock file on greenkeeper
package update PRs.

- While cross-origin PRs don't have access to project secrets,
  installing and executing an external module in a step that has push
  access to the repo on GitHub seemed risky
- yarn generate-lock-file does not fully resolve the modules and
  confusingly generates a differently formatted yarn.lock file
  [yarnpkg/yarn#2340][issue]
- Run a full yarn install so that a expected yarn.lock is
  generated/updated

[issue]: https://github.com/yarnpkg/yarn/issues/2340
[module]:https://github.com/clarkbw/circleci-update-yarn-lock

### DIFF
--- a/bin/circleci-update-yarn-lock
+++ b/bin/circleci-update-yarn-lock
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+git log --name-status HEAD^..HEAD | grep "fix(package): update" || { echo ">> not an update commit"; exit 0; }
+
+yarn install --no-progress --pure-lockfile
+
+yarn check && { echo ">> yarn check passed, yarn.lock does not need to update"; exit 0; }
+
+yarn install --force --no-progress
+
+name="greenkeeper[bot]"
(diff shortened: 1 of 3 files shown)


bin/circleci-update-yarn-lock
```diff
@@ -0,0 +1,25 @@
...
+echo ">> running yarn install --force"
+yarn install --force --no-progress
+
+name="greenkeeper[bot]"
+email="greenkeeper[bot]@users.noreply.github.com"
+
+git config user.name "$name"
+git config user.email "$email"
+
+echo ">> git config $name <$email>"
```
wfleming: nit: avoid var interpolation in `echo`. Use `printf` here if you want to log these (though since they're hard-coded I also personally think you could just drop this message.

PR #229, opened 2017-03-21
Prevent libs from messing with stdout

filipesperandio:
Add back "Prevent errors from unsupported plugins/rules/modules" once more


filipesperandio:
Analysis is failing on `test/integration` files. Thought [this](https://github.com/codeclimate/codeclimate-eslint/blob/master/.codeclimate.yml#L17) would prevent that.

dblandin:
> Analysis is failing on test/integration files. Thought this would prevent that.

"test/**"

@filipesperandio That would cover any files under the `test/` directory but the Code Climate issue seems to be targeting a file under the `integration/` directory, which is not covered by that pattern.

filipesperandio:
@dblandin Ah! Right!

dblandin:
@filipesperandio Could you rebase this branch with origin/master to address the merge conflicts?

Happy to review once GitHub is happy.

bin/eslint.js
```diff
@@ -4,4 +4,8 @@ const CODE_DIR = "/code";
 process.chdir(CODE_DIR);
 
 const ESLint = require("../lib/eslint");
-ESLint.run({ dir: CODE_DIR });
+const issues = ESLint.run({ dir: CODE_DIR });
```
filipesperandio: @wfleming There is a major difference here from the previous behaviour, as it now accumulates all issues before printing them all. I hope it is OK.

I see this could be potentially problematic on large repos and if the engine is actually making use of the output stream  (which I don't have context yet).

wfleming: > There is a major difference here from the previous behaviour, as it now accumulates all issues before printing them all. I hope it is OK.

It is very much not preferred. There can be significant performance penalties for buffering this way.

Was there no reasonable alternative?

filipesperandio: It makes testing a breeze!

filipesperandio: I think this is a good solution: https://github.com/codeclimate/codeclimate-eslint/pull/230

wfleming: Talked this through with @GordonDiggs IRL, and I'm afraid that would be a showstopper in this case. At points in the past we did do more buffering of issues in this engine, and large repos with many issues could bring the engine to its knees.

As I was writing this, I see it looks like you already addressed this with #230, though, so that's great. Thanks!

filipesperandio:
@dblandin I think this is ready to re-revert (again).

PR #230, opened 2017-03-21
Keep it testable with no performance penalty

filipesperandio:
@wfleming this seems to be a good alternative, testable and performant


filipesperandio:
Merging into the other branch...

Issue #222, opened 2017-03-20
Ignore ESLint rules that rely on module resolution


dblandin:
The ESLint Code Climate engine is intended to run statically against a project workspace. Project dependencies [are not installed via npm or yarn prior to analysis](https://github.com/codeclimate/codeclimate/issues/480). Any ESLint rules that require module resolution will result in a Code Climate issue similar to the following:

```json
  {
    "type": "issue",
    "categories": [
      "Style"
    ],
    "check_name": "import/no-unresolved",
    "description": "Unable to resolve path to module 'react'.",
    "content": {
      "body": "For more information visit Source: http://eslint.org/docs/rules/\n"
    },
    "location": {
      "path": "index.js",
      "positions": {
        "begin": {
          "line": 1,
          "column": 19
        },
        "end": {
          "line": 1,
          "column": 19
        }
      }
    },
    "remediation_points": 50000,
    "fingerprint": "26297b57d7792d3a5d99bbec8b9e94e0",
    "engine_name": "eslint"
  }
```

We should either maintain a blocklist of rules to skip during analysis or implement a patch that triggers a skip whenever an unexpected module resolution is executed. We may also want to warn to STDERR when ignoring an ESLint check, providing the check name in the message.

If we decide upon a blocklist approach, all currently vendored plugins should be audited for rules that rely on module resolution.

dblandin:
Closing with #231 🎉

PR #231, opened 2017-03-22
Rule Blocklist

pointlessone:
Add a way to disable rules.

One particular case is #222.

### DIFF
--- a/lib/config_finder.js
+++ b/lib/config_finder.js
@@ -1,0 +1,28 @@
+"use strict";
+
+const CONFIG_FILES = require("eslint/lib/config/config-file").CONFIG_FILES
+  , FileFinder = require("eslint/lib/file-finder")
+  , path = require("path");
+
+module.exports = function(configFile, analysisFiles) {
+  if (configFile) {
+    return [configFile];
+  }
+
+  const dirs = analysisFiles.map(function(fileName){
+    return path.dirname(fileName);
+  });
+  const directories = Array.from(new Set(dirs));
+
+  const localConfigFinder = new FileFinder(CONFIG_FILES, process.cwd());
+
+  const configs = new Set();
+  for (const dir of directories) {
+    const configFiles = localConfigFinder.findAllInDirectoryAndParents(dir);
+    for (const file of configFiles) {
+      configs.add(file);
+    }
+  }
+
+  return Array.from(configs);
+};
--- a/lib/eslint-patch.js
+++ b/lib/eslint-patch.js
@@ -7,6 +7,7 @@ const ConfigFile = require("eslint/lib/config/config-file");
 
 const Config = require("eslint/lib/config");
 const ConfigUpgrader = require("./config_upgrader");
+const RuleBlocklist = require("./rule_blocklist");
 
 module.exports = function patch() {
   const skippedModules = [];
@@ -43,8 +44,11 @@ module.exports = function patch() {
   Config.prototype.getConfig = function(filePath) {
     const originalConfig = originalGetConfig.apply(this, [filePath]);
     const configUpgrader = new ConfigUpgrader();
+    const ruleBlocklist = new RuleBlocklist();
 
-    return configUpgrader.upgrade(originalConfig);
+    return ruleBlocklist.filter(
+      configUpgrader.upgrade(originalConfig)
+    );
   };
 
   return require('eslint');
--- a/lib/rule_blocklist.js
+++ b/lib/rule_blocklist.js
@@ -1,0 +1,86 @@
+"use strict";
+
+const Config = require("eslint/lib/config")
+  , merge = require("eslint/lib/config/config-ops").merge;
+
+const blocklistedRules = [
+  "import/no-unresolved",
+  "import/extensions",
+  "import/no-absolute-path"
+];
+
+function filterRules(rules) {
+  let report = [];
+
+  for (const name of blocklistedRules) {
+    if (Reflect.has(rules, name)) {
+      let config = rules[name];
+      if (config.constructor !== Array) {
+        config = [config];
+      }
+      let severity = config.shift();
+
+      if (severity !== 0 && !(typeof severity === "string" && severity.toLowerCase() === "off")) {
+        rules[name] = "off";
+        report.push(`* ${name}`);
+      }
+    }
+  }
+  return report;
+}
+
+class RuleBlocklist {
+  constructor() {
+    this._report = [];
+  }
+
+  filter(originalConfig) {
+    if (typeof originalConfig === "undefined" || originalConfig === null) {
+      return {};
+    }
+
+    let config = merge({}, originalConfig);
+
+    this._report = [];
+
+    if (Reflect.has(config, "rules")) {
+      let report = filterRules(config.rules);
+      this._report = this._report.concat(report);
+    }
+
+    return config;
+  }
+
+  get report() {
+    return [].concat(this._report);
+  }
+
+  static report(configs) {
+    const reports = configs.map(function(configFile) {
+      let report = [];
+
+      const blocklist = new RuleBlocklist();
+      const config = new Config({
+        configFile: configFile,
+        cwd: process.cwd()
+      });
+      blocklist.filter(config.useSpecificConfig);
+
+      if (report.length > 0 || blocklist.report.length > 0) {
+        report = report.concat(blocklist.report);
+      }
+
+      return report;
+    });
+
+    if (reports.length === 0) {
+      return [];
+    } else {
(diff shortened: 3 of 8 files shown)


lib/eslint-patch.js
```diff
@@ -7,6 +7,7 @@ const ConfigFile = require("eslint/lib/config/config-file");
 
 const Config = require("eslint/lib/config");
 const ConfigUpgrader = require("./config_upgrader");
+const RuleBlacklist = require("./rule_blacklist");
```
pbrisbin: Could you rename your files and variables to use "blocklist" throughout?

lib/rule_blocklist.js
```diff
@@ -0,0 +1,84 @@
...
+
+const blocklistedRules = {
+  "no-unresolved": "Code Climate doesn't fetch your project's dependencies so this rule produces high amount of false positives."
+};
+
+function filterRules(rules) {
+  let report = [];
+
+  for (const name in blocklistedRules) {
+    if (Reflect.has(rules, name)) {
```
pbrisbin: Just curious, what's `Reflect`? I don't see it being `require`d.

lib/config_finder.js
```diff
@@ -0,0 +1,30 @@
+"use strict";
+
+const
+    CONFIG_FILES = require("eslint/lib/config/config-file").CONFIG_FILES
```
pbrisbin: :lipstick: not sure what project style is, but elsewhere you put the first value on the line with `const`.

lib/rule_blocklist.js
```diff
@@ -0,0 +1,84 @@
+"use strict";
+
+const a = "a"
```
pbrisbin: What's `a` for?

dblandin:
This is looking good! The current rule configured in the blocklist is missing the `import/` prefix, but after fixing that the engine successfully ignored the rule. 🎉 

Instead of 

```
Ignored rules:

* import/no-unresolved:
  Code Climate doesn't fetch your project's dependencies so this rule produces high amount of false positives.
```

what do you think of outputting a list of ignored rules (if any) after a brief explanation:

```
Ignoring the following rules that rely on module resolution:

* import/no-unresolved
```

lib/rule_blocklist.js
```diff
@@ -0,0 +1,84 @@
...
+  , Config = require("eslint/lib/config")
+  , merge = require("eslint/lib/config/config-ops").merge
+  ;
+
+const blocklistedRules = {
+  "no-unresolved": "Code Climate doesn't fetch your project's dependencies so this rule produces high amount of false positives."
+};
```
dblandin: Mentioned in another comment but repeating here inline: I don't think we need to support separate explanations for blocklisted rules just yet. This can be an array of rules and we can provide an explanation after updating the config.

`no-resolved` should be `import/no-resolved`.

pointlessone: > ```
> Ignoring the following rules that rely on module resolution:
> 
> * import/no-unresolved
> ```

Right now we have just one reason (we know of) to turn off rules. And we have just one rule (I didn't do extensive research on this yet). It's an appropriate way to format the message for it.

However, it doesn't seem to be the only possible reason. I specifically chosen this approach because there might be others and it will be a bit easier to change it in the future — just add it to this object, no code changes needed.

Here are a few examples of other reasons we may want to disable a rule I can think of:
* A platform-specific rule that is incompatible with Linux
* A rule that relies on an external tool that is not in the container
* A rule that tries to use network
* A rule that is incompatible with Node version we run

I'm not opposing the proposed change. I'm just presenting my reasoning behind the design. I'll make the change if you prefer it.

lib/rule_blocklist.js
pointlessone: It's a [standard built-in object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect).

lib/rule_blocklist.js
dblandin: We should definitely audit all rules currently available via our vendored plugins and add any that rely on module resolution to this blocklist either in this PR or the next one.

Here's one example of another rule that should be in the blocklist:
`import/extensions`
https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/extensions.js

> Here are a few examples of other reasons we may want to disable a rule I can think of

Until we actually want to ignore rules for another reason, I'd stick with a simpler implementation that doesn't map particular rules to warning messages. I don't think it's worth adding now. If we need it later, we can add it then.

dblandin:
@pointlessone Anything I can do to help push this along?

efueger:
Another rule that should probably be included in the blocklist:
- [import/no-absolute-path](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-absolute-path.md)

Otherwise, repos see issues like: "Resolve error: unable to load resolver "babel-module"."

dblandin:
@pointlessone Looks like a good portion of the diff is refactoring the config finder. Would you mind splitting that refactor into a separate commit so the rule blocklist additions can stand alone?

dblandin:
@pointlessone Have you been able to audit the currently vendored plugins for other rules that rely on module resolution? It would be great if we could catch and blocklist any others.

pointlessone:
@dblandin No, I haven't audited other rules. It's a rather time consuming task. I'm not sure if that's an effective use of time. I thought we could push this and add more rules as we hit more false positives. /cc @GordonDiggs 

TBH, I'm not sure about `import/extensions` rule either. According to [docs](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md) it seem only deal with extensions and shouldn't rely on module resolution. On the other hand, implementation seem to use `resolve`. ¯\\\_(ツ)\_/¯

dblandin:
I'm fine rolling with this if you think an audit is too time intensive right right now.

I was thinking of a simple ack/grep for rules calling out to `resolve` and checking each briefly for module resolution:

```
$ ack "resolve" ./**/rules/**/*.js | grep ".js$"

eslint-config-airbnb-base/rules/imports.js
eslint-config-airbnb-base/rules/style.js
eslint-config-airbnb/rules/react.js
eslint-config-simplifield/node_modules/eslint-plugin-angular/rules/deferred.js
eslint/lib/rules/global-require.js
eslint/lib/rules/no-alert.js
eslint/lib/rules/no-loop-func.js
eslint/lib/rules/no-path-concat.js
eslint/lib/rules/no-unmodified-loop-condition.js
eslint/lib/rules/no-unused-vars.js
eslint/lib/rules/no-use-before-define.js
eslint/lib/rules/prefer-const.js
eslint/lib/rules/prefer-promise-reject-errors.js
eslint-plugin-angular/rules/deferred.js
eslint-plugin-filenames/lib/rules/match-exported.js
eslint-plugin-filenames/lib/rules/match-regex.js
eslint-plugin-filenames/lib/rules/no-index.js
eslint-plugin-flowtype/dist/rules/defineFlowType.js
eslint-plugin-import/lib/rules/extensions.js
eslint-plugin-import/lib/rules/no-duplicates.js
eslint-plugin-import/lib/rules/no-internal-modules.js
eslint-plugin-import/lib/rules/no-restricted-paths.js
eslint-plugin-import/lib/rules/no-unresolved.js
eslint-plugin-inferno/lib/rules/no-unused-prop-types.js
eslint-plugin-inferno/lib/rules/prop-types.js
eslint-plugin-jsx-a11y/lib/rules/aria-proptypes.js
eslint-plugin-jsx-a11y/src/rules/aria-proptypes.js
eslint-plugin-node/lib/rules/no-hide-core-modules.js
eslint-plugin-node/lib/rules/no-unpublished-bin.js
eslint-plugin-node/lib/rules/shebang.js
eslint-plugin-promise/rules/lib/is-promise.js
eslint-plugin-promise/rules/no-return-wrap.js
eslint-plugin-promise/rules/param-names.js
eslint-plugin-react/lib/rules/no-unused-prop-types.js
eslint-plugin-react/lib/rules/prop-types.js
eslint-plugin-react/lib/rules/require-default-props.js
```

pointlessone:
@dblandin Split the commits.

Simple grep is a rather naive approach. I imagine it can hit all sorts of unrelated things: comments, any function/var that has this rather generic term in them, etc.

Moreover, not every `resolve` is destined to hit external modules. After all there's some code in the project.

Let's merge this PR as is (or with the changes directly related to it) and expand the list of blocked rules in a separate PR. I'll go through the list you posted and add rules if they primarily deal with external modules. Otherwise, users can mark false-positives.

dblandin:
Sounds good!

PR #241, opened 2017-04-03
Fix boolean config flags behavior

wfleming:
Due to some oddness with our `cc-yaml` parser right now, all keys under `config` for an engine are parsed as strings. The engine did not take this into account when inspecting boolean config options `debug` and `ignore_warnings`, which basically meant explicitly specifying *any* value for these flags was always interpreted as `true`, and the only way to get `false` was to not specify the flag at all (since both values default to `false`). (Or you could specify the value as `""` in YAML, since `!!""` is `false` in JS, but that's a dumb hack.)

I started here by writing some red tests & extracting the config behavior out of `lib/eslint.js`, then fixing the bug there to see tests change.

I'll squash up commits before merging.

### DIFF
--- a/lib/engine_config.js
+++ b/lib/engine_config.js
@@ -8,21 +8,30 @@ EngineConfig.prototype.includePaths = function() {
   return this.engineJSON.include_paths;
 };
 
-EngineConfig.prototype.userConfig = function() {
-  return new UserEngineConfig(this.engineJSON.config || {});
+// cc-yaml currently ends up passing bools as stringy equivalents
+function _coerceBool(val) {
+  if (typeof(val) === "string") {
+    return val === "true";
+  } else {
+    return !!val;
+  }
 };
 
 function UserEngineConfig(json) {
   this.json = json;
 };
 
+EngineConfig.prototype.userConfig = function() {
+  return new UserEngineConfig(this.engineJSON.config || {});
+};
+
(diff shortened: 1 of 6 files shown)


lib/engine_config.js
```diff
@@ -0,0 +1,48 @@
...
+function EngineConfig(path) {
+  this.engineJSON = JSON.parse(fs.readFileSync(path));
+};
+
+EngineConfig.prototype.includePaths = function() {
+  return this.engineJSON.include_paths;
+};
+
+// cc-yaml currently ends up passing bools as stringy equivalents
+function _coerceBool(val) {
```
chrishulton: jw, why not define this as a function on say the EngineConfig prototype?

wfleming:
> The tests seem to capture the intended behavior so LGTM. I guess this wouldn't support something like "ignore_warnings": 1 but if we clearly state you should only use true for a value, this should be fine.

Yeah, this is really more intended to work around or own YAML parsing weirdness than anything else. Docs say users should set it as a bool, so I'm fine not supporting odd edge cases for that.

lib/engine_config.js
wfleming: Just because it's private, and putting it on the prototype makes it "private by convention" (since it's technically still exposed to outside modules), as opposed to actually really private.

zenspider:
Yay!

PR #253, opened 2017-04-11
Add support for new ESLint plugins

leftees:
This pull request adds support for the following plugins:

- ~eslint-plugin-html~ (this is already here, but maybe it wasn't clear how to use it because we ignore html files unless the user explicitly asks us not to)

- eslint-plugin-ejs

- eslint-plugin-json


maxjacobson:
@leftees while QA-ing this change, I learned about an under-documented feature of this engine, so I added a note to the README on this branch. WDYT?

skye2k2:
JMTC, but until checking this repository, I was unaware that eslint-plugin-html was already incorporated, due to it being omitted from the docs (https://docs.codeclimate.com/docs/eslint#section-supported-plugins). Adding it could help others like me, who didn't think to read the source code. Thanks!

maxjacobson:
@skye2k2 yep, thanks for the reminder. We try to keep that list up-to-date, but it looks like this one fell thru the cracks. We'll update it. Today I learned "JMTC" as well :)

PR #257, opened 2017-04-24
Update auto-yarn strategy

maxjacobson:
Sometimes the commit messages look like "chore(package)", not
"fix(package)", but we should still run yarn in those cases.

Example: https://github.com/codeclimate/codeclimate-eslint/pull/256

I'm guessing it's just a chore when it's a devDependency.

### DIFF
--- a/bin/circleci-update-yarn-lock
+++ b/bin/circleci-update-yarn-lock
@@ -16,5 +16,5 @@ email="greenkeeper[bot]@users.noreply.github.com"
 git config user.name "$name"
 git config user.email "$email"
 
-git commit --all --message "chore(package): update yarn.lock"
+git commit --all --message "update yarn.lock"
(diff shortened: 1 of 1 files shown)


dblandin:
Would this cause an infinite loop of build + commit cycles unless we change the commit message we use when updating the `yarn.lock` file?

Right now that commit message is:
```
chore(package): update yarn.lock
```

which would match `(package): update`.

maxjacobson:
Ah, good point. I didn't notice that part. I'm inclined to change that message, then. Maybe just `update yarn.lock`. I think we'll generally merge greenkeeper PRs with squash-and-merge and discard this message anyway. WDYT?

dblandin:
> Maybe just update yarn.lock

Sounds good to me!

PR #258, opened 2017-04-24
Fix `make test` and `make integration`

maxjacobson:
We're hitting errors where the `mocha` command is not available. I
believe this is because we were running `yarn` in `/usr/src/app`, and
then trying to run tests in `/code`, where the node_modules aren't
available.

For a sec, I assumed this was related to
https://github.com/codeclimate/codeclimate-eslint/pull/256 but I
confirmed the issue predates that change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,14 +16,12 @@ image:
 
 integration: image
 	docker run -ti --rm \
-		--volume $(PWD):/code \
-		--workdir /code \
+		--workdir /usr/src/app \
 		$(IMAGE_NAME) npm run $(NPM_INTEGRATION_TARGET)
 
 test: image
 	docker run -ti --rm \
-		--volume $(PWD):/code \
-		--workdir /code \
+		--workdir /usr/src/app \
 		$(IMAGE_NAME) npm run $(NPM_TEST_TARGET)
 
 citest:


dblandin:
Might want to ping @filipesperandio for some context.

maxjacobson:
Sounds good, I'll update the reviewer to @filipesperandio

filipesperandio:
Though this will solve the missing dependencies problem, we are missing a way to update `yarn.lock` with new dependencies, as they are resolved only inside the container.

maxjacobson:
@filipesperandio the way to update `yarn.lock` on the local filesystem is to run commands like `bin/yarn add foo`, `bin/yarn remove foo`, `bin/yarn upgrade`, etc

This is mentioned [here](https://github.com/codeclimate/codeclimate-eslint/blob/master/CONTRIBUTING.md#adding-and-upgrading-plugins) although I'm not sure how discoverable that file is

filipesperandio:
@maxjacobson True. I found that after my comment. I think it is still too apart from the build process. For example, if `package.json` is edited directly, there nothing that enforces `yarn` to run again. I will make a PR with what I have in mind.

maxjacobson:
@filipesperandio cool, looking forward to seeing that. Going to merge this in the meantime

PR #260, opened 2017-04-24
Build improvements

filipesperandio:
The idea is to make sure `yarn.lock` gets updated during development in case `package.json` is manually edited.

The dependencies chain in Makefile seems ok, but this changes might turn the flow a bit convoluted as the result
of `make yarn.lock` generate files which invalidates `docker build` cache.

I think there is also room for improving how we generate `lib/docs` and `engine.json` through `Dockerfile` without having
this process to indirectly dependend on `package.json` (`jq > npm install > package.json`)

Please, let me know if this is worth keeping.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,16 @@ WORKDIR /usr/src/app
 COPY bin/docs ./bin/docs
 COPY engine.json package.json yarn.lock ./
 
+RUN mkdir /usr/local/node_modules
+ENV PREFIX=/usr/local/node_modules
+ENV PATH=$PREFIX/.bin:$PATH
+ENV NODE_PATH=$PREFIX
+ENV NPM_CONFIG_PREFIX=$PREFIX
+
 RUN apt-get install -y git jq yarn && \
-    yarn install && \
-    version="v$(npm -j ls eslint | jq -r .dependencies.eslint.version)" && \
+    yarn config set prefix $PREFIX && \
+    yarn install --modules-folder $PREFIX && \
+    version="v$(yarn list eslint | grep eslint | sed -n 's/.*@//p')" && \
     bin/docs "$version" && \
     cat engine.json | jq ".version = \"$version\"" > /engine.json && \
     apt-get purge -y git jq yarn && \
@@ -20,6 +27,7 @@ RUN apt-get install -y git jq yarn && \
 RUN adduser --uid 9000 --gecos "" --disabled-password app
 COPY . ./
 RUN chown -R app:app ./
+RUN chown -R app:app $PREFIX
 
 USER app
 
--- a/Makefile
+++ b/Makefile
@@ -14,13 +14,15 @@ endif
 image:
 	docker build --rm -t $(IMAGE_NAME) .
 
-integration: image
+integration: yarn.lock
 	docker run -ti --rm \
+		-v $(PWD):/usr/src/app \
 		--workdir /usr/src/app \
 		$(IMAGE_NAME) npm run $(NPM_INTEGRATION_TARGET)
 
-test: image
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -23,7 +23,4 @@ RUN chown -R app:app ./
 
 USER app
 
-VOLUME /code
-WORKDIR /code
```
pbrisbin: IIRC this is a requirement of the SPEC. I'm not sure we're in love with that requirement, or if _all_ our engines follow it, but while it's there we might as well not make changes that invalidate engines that _are_ currently in compliance.

filipesperandio: True. Someone mentioned that to me before, I forgot and thought I had added those.
What about moving everything from `/usr/src/app` to `/code`, though?

filipesperandio:
@maxjacobson things are not there yet, like what I was aiming for to really speed up test execution.

For us to actually get `make` benefits, we would have to avoid using some `PHONY` targets (or stop depend on them). The way things are currently structured it is a bit difficult to accomplish that.

But on the case of `yarn.lock`, this target depends on `package.json`, if nothing changes, it does not have to run again. `image` is also a dependency and it is `PHONY` so it will always run :(

Example:
1. You `touch package.json` and run `make test`:
`test` depends on `yarn.lock` which depends on `package.json` (which changed) and `image` (PHONY).
All tasks run: `image`, `yarn.lock`, `test`.

2. Nothing changes on `package.json` and you run `make test`:
`test` depends on `yarn.lock` which depends on `package.json` (which didn't change) and `image`.
Tasks that run: `image`, `test`. 
Since `package.json` wasn't changed, `make` knows there is nothing to do for `yarn.lock`.

maxjacobson:
Cool, thanks for explaining. If you can think of a way to achieve those goals, that would be great.

I think I'm in favor of making this simple change to address your original concern about keeping `yarn.lock` up-to-date in the meantime:

```diff
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ test: image
                $(IMAGE_NAME) npm run $(NPM_TEST_TARGET)

 citest:
+       bin/yarn check && \
        docker run --rm \
                --workdir /usr/src/app \
                $(IMAGE_NAME) sh -c "npm run test && npm run integration"
```

Dockerfile
filipesperandio: Gotcha.

filipesperandio:
@maxjacobson Running some experiments to see if I can get closer to that goal...
The fact that `yarn install` add `node_modules` dir to the working dir makes things a bit weird, it exists after `docker build` but it is not there if you override the volume... and then mixing `npm` and `yarn` commands makes it even more difficult to change... :(

maxjacobson:
🤷‍♂️ yeah, I dunno. IMO the status quo is kind of fine for now.

filipesperandio:
Not happy yet. I had some progress early this week but need to be polished that a bit before I push.

filipesperandio:
I think I manage to get the result I was looking for. In my laptop `make test` now runs in `3.8s` instead of `10s` if the image is up-to-date. Any changes to `package.json` or `Dockerfile` will trigger `make image` but if nothing change in those files, there is no need to rebuild every time.

filipesperandio:
@maxjacobson One last thing: I might have messed a bit with the `eslint` version lookup in this movement of `node_modules` out of the local directory:
- `npm list` is not working as before;
- `npm info` sounded as a replacement but it looks at the registry instead of the resolved dependencies
- I just added a `yarn` version of the original command, but the output format needed some tuning.

maxjacobson:
I see. This still LGTM

PR #262, opened 2017-04-25
Check yarn.lock's correctness

maxjacobson:
I found myself suggesting this in another PR, so I decided to just open it as its own PR https://github.com/codeclimate/codeclimate-eslint/pull/260#issuecomment-297179681

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,4 +27,4 @@ test: image
 citest:
 	docker run --rm \
 		--workdir /usr/src/app \
-		$(IMAGE_NAME) sh -c "npm run test && npm run integration"
+		$(IMAGE_NAME) sh -c "yarn check && npm run test && npm run integration"
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "eslint-plugin-node": "^4.2.1",
     "eslint-plugin-prettier": "^2.0.1",
     "eslint-plugin-promise": "^3.5.0",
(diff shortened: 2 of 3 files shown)


maxjacobson:
Hm, maybe this won't work actually -- we need builds to pass in order for greenkeeper PRs to run the `bin/circleci-update-yarn-lock` script. Will think more about that later, but shouldn't merge this for now ⚠️

filipesperandio:
@maxjacobson Yeah... I was giving some thought to this as well, yesterday.
I think keeping `yarn.lock` sane is a task for the developer, meaning I would rather have it on its own target or as `test` dependency instead of `citest`, WDYT?

filipesperandio:
@maxjacobson Was looking into `bin/circleci-update-yarn-lock` and it also does `yarn check`... maybe that's good enough?

maxjacobson:
@filipesperandio that's triggered when greenkeeper opens a PR updating `package.json`. It only runs on `greenkeeper/*` branches (https://github.com/codeclimate/codeclimate-eslint/blob/da4d8ed7ac400b4ff4066e36644de0509e5e28cf/circle.yml#L34-L37)

filipesperandio:
Does `greenkeeper` check dependencies once a day or so? (Not too familiar with greenkeeper)

maxjacobson:
Something like that. They make one PR per dependency package, and I think they sometimes open multiple PRs in the same day.

PR #265, opened 2017-05-04
Bumps eslint-plugin-react 6.10.0 -> 6.10.3

chrishulton:
In particular this incorporates the `Fix jsx-indent crash` bug fix.
https://github.com/yannickcr/eslint-plugin-react/releases/tag/v6.10.1


chrishulton:
@maxjacobson fyi actually looking at the versioning we use here, `"^6.10.0"` is going to bump it to `6.10.3` so i'm going to explicitly put that in the package.json.

maxjacobson:
OK still looks good to me

PR #279, opened 2017-05-25
Add support for eslint-plugin-ember

kevinansfield:
[eslint-plugin-ember](https://github.com/netguru/eslint-plugin-ember) is used to enforce best practices more than syntax rules which is handled via `eslint-plugin-ember-suave`

PR #303, opened 2017-07-11
Add support for eslint-plugin-ember

amiel:
eslint-plugin-ember can be found here: https://github.com/netguru/eslint-plugin-ember

I have tested this plugin by following the directions at https://github.com/codeclimate/codeclimate-eslint/blob/master/CONTRIBUTING.md and found it to work with our .codeclimate.yml configuration.

FYI: I discussed adding this in a zendesk customer support conversation with @davehenton.

I appreciate you looking in to adding this plugin as it will fix our codeclimate build. Temporarily, I have added a specific .eslintrc-codeclimate.js for codeclimate to use that does not reference this plugin, but we are now having issues with keeping the extra configuration in sync and have found that specifying a specific eslintrc bypasses the automatic use of per-directory eslintrc files, which causes false positives codeclimate. Anyway, it would be really great for our team to get this merged :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "eslint-plugin-angular": "^2.0.0",
     "eslint-plugin-babel": "^4.1.1",
     "eslint-plugin-ejs": "^0.0.2",
+    "eslint-plugin-ember": "^3.6.1",
     "eslint-plugin-ember-suave": "^1.0.0",
     "eslint-plugin-filenames": "^1.1.0",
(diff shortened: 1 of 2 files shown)


aureliosaraiva:
https://github.com/codeclimate/codeclimate-eslint/pull/305

amiel:
Closed in favor of #279.

akatov:
can you please upvote #306 - update to eslint-plugin-ember?

PR #308, opened 2017-07-23
ESLint 4

### LINKED ISSUE #283: Support for ESLint 4.0.0

pointlessone:
Fixes #283


lib/config_upgrader.js
```diff
@@ -235,8 +236,9 @@ class ConfigUpgrader {
       const config = new Config({
         configFile: configFile,
         cwd: process.cwd()
-      });
-      upgrader.upgrade(config.useSpecificConfig);
+      },
+      new Linter());
```
maxjacobson: is this indentation correct? It took me a moment, reading it, to realize that this value is an argument to `new Config`

lib/config_upgrader.js
```diff
@@ -235,8 +236,9 @@ class ConfigUpgrader {
       const config = new Config({
         configFile: configFile,
         cwd: process.cwd()
-      });
-      upgrader.upgrade(config.useSpecificConfig);
```
maxjacobson: will you please summarize why is it we don't seem to use the `ConfigUpgrader` outside of tests anymore? If we don't need it, we should delete it totally.

pointlessone: Agree, it is confusing. Closing brace is 2 spaces outdented. In that it is correct. But the second argument is level with outer function call. In that it is incorrect. Probably, have to move the opening brace to a new line and indent all arguments.

lib/config_upgrader.js
pointlessone: Gordon told me that we can skip it. Release without upgrader in a separate channel for people who really want to use it and prepared to manually upgrade their configs. Add it later when we're ready to promote it to `latest`.

But I'm not sure about that now. ESLint 4 is much stricter about configs. It validates all keys, throws on anything unknown/dropped in previous versions. So maybe it's more urgent than previously thought.

 There's another bug to fix (plugin loading) before this can be considered working with a perfect config, but it would still fail integration tests because they all have invalid configs fro ESLint 4.

README.md
```diff
@@ -56,6 +56,94 @@ eslint:
...
+* eslint-plugin-lodash-fp
+* @br/eslint-plugin-laws-of-the-game
+* eslint-plugin-mongodb
+* eslint-plugin-prettier
+* eslint-plugin-promise
+* eslint-plugin-react-intl
+* eslint-plugin-security
+* eslint-plugin-standard
+* eslint-plugin-vue
+* eslint-plugin-xogroup
```
maxjacobson: This detail is great, but I worry that most people won't see it because they'll just look at the README for the default branch. Will you make an additional PR updating the readme in the default branch clarifying that that branch is for eslint-3, and if you're looking for eslint-4 to click over to the other page?

pointlessone: Great idea! Will do.

Do you think I should also mention the incompatibility there? So that people knew they might not be able to upgrade just yet.

maxjacobson: I think that's reasonable, yea

pointlessone: #312

PR #315, opened 2017-08-04
Fix Dockerfile

pointlessone:
This fixes an issue when APT index was cached for longer than needed which lead to missing packages and as result docker was unable to finish building the image.

Please let me know if I should provide a detailed explanation for why this is better than it was.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,35 @@
 FROM node:6.10.0-slim
 LABEL maintainer "Code Climate <hello@codeclimate.com>"
 
-RUN apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg && \
-    echo "deb http://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
-    apt-get update
+RUN adduser --uid 9000 --gecos "" --disabled-password app
 
 WORKDIR /usr/src/app
-COPY bin/docs ./bin/docs
-COPY engine.json package.json yarn.lock ./
 
-RUN mkdir /usr/local/node_modules
 ENV PREFIX=/usr/local/node_modules
 ENV PATH=$PREFIX/.bin:$PATH
 ENV NODE_PATH=$PREFIX
 ENV NPM_CONFIG_PREFIX=$PREFIX
 
-RUN apt-get install -y git jq yarn && \
+RUN mkdir $PREFIX
+
+COPY bin/docs ./bin/docs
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -24,12 +28,10 @@ RUN apt-get install -y git jq yarn && \
     apt-get purge -y git jq yarn && \
     apt-get autoremove --yes
 
-RUN adduser --uid 9000 --gecos "" --disabled-password app
-COPY . ./
-RUN chown -R app:app ./
 RUN chown -R app:app $PREFIX
 
 USER app
+COPY . ./
```
dblandin: Can you explain the reordering of the copy and chown lines? I think we might still want to copy and change permissions as root before dropping down to the `app` user.

pointlessone: You're probably right. The motivation was to move rarely changing layers closer to the top so that they were not blown up for no good reason.

Dockerfile
```diff
@@ -1,33 +1,36 @@
...
 
-RUN apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg && \
-    echo "deb http://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
-    apt-get update
+RUN adduser --uid 9000 --gecos "" --disabled-password app
 
 WORKDIR /usr/src/app
-COPY bin/docs ./bin/docs
-COPY engine.json package.json yarn.lock ./
+RUN chown -R app:app ./
```
filipesperandio: Don't think we need this. Do we?

Dockerfile
```diff
@@ -1,33 +1,35 @@
...
-COPY engine.json package.json yarn.lock ./
 
-RUN mkdir /usr/local/node_modules
 ENV PREFIX=/usr/local/node_modules
 ENV PATH=$PREFIX/.bin:$PATH
 ENV NODE_PATH=$PREFIX
 ENV NPM_CONFIG_PREFIX=$PREFIX
 
-RUN apt-get install -y git jq yarn && \
+RUN mkdir $PREFIX
```
dblandin: One more thing: should we move this `mkdir` to the `RUN` below just before the folder is used to remove one more docker layer?

pointlessone: We can. Though, this particular layer is 0 bytes (added to the final image) and cached all the time. Don't see much downside keeping it.

dblandin: If we don't need the layer, I'm fine with dropping it. Not sure if yarn will automatically create it for you.

PR #321, opened 2017-08-21
Create issue_template.md

efueger:
Move oss engine requests to oss projects (instead of handling internally)

### DIFF
--- a/issue_template.md
+++ b/issue_template.md
@@ -6,9 +6,12 @@
 
 ## Steps to reproduce
 
+> _Include configurations used in your: <br>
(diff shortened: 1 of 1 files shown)


issue_template.md
```diff
@@ -0,0 +1,14 @@
+# Bug Template
+
+## Actual Behavior
+
+## Expected Behavior
+
+## Steps to reproduce
+
+## Relevant links
```
GordonDiggs: Maybe worth adding "What channel of the engine are you using"?

PR #326, opened 2017-09-17
Upgrade dependencies

benmccann:


dblandin:
@benmccann Thanks!

benmccann:
Thanks Devon! Do I need to wait some amount of time before this change will be available on the Code Climate servers?

dblandin:
Shouldn't take too long but yes, I'll need to push the update through our systems before it's available on codeclimate.com. I'll ping you here when that's done 👍

dblandin:
@benmccann Should be all set!

benmccann:
Thank you!!

PR #329, opened 2017-09-19
Update eslint-plugin-ember to 4.5.0

kculafic:
Upgrade `eslint-plugin-ember` to version 4.5.0 in `package.json`


kculafic:
@dblandin Oof, it was not. Good catch

PR #342, opened 2017-10-12
Add more packages and blocklist more rules

GordonDiggs:


GordonDiggs:
I tested this on some OSS repos and it seems to work well

package.json
```diff
@@ -15,23 +15,39 @@
...
     "eslint-config-standard": "^10.0.0",
     "eslint-config-standard-jsx": "^4.0.0",
     "eslint-config-standard-react": "^5.0.0",
     "eslint-config-strongloop": "^2.1.0",
+    "eslint-config-vue": "^2.0.2",
+    "eslint-config-xo": "^0.18.2",
+    "eslint-config-xo-react": "^0.13.0",
+    "eslint-config-xo-space": "^0.16.0",
+    "eslint-import-resolver-node": "^0.3.1",
+    "eslint-import-resolver-webpack": "^0.8.3",
```
dblandin: We've had trouble with these revolvers in the past. Do we need to include them if we're blocklisting some of the `import/` rules?

GordonDiggs: I think we do because the blocklisting happens when we report the issues. If we don't have these here, the analysis will error working with the config, right?

dblandin: Ahh, probably... sounds good to me.

Issue #320, opened 2017-08-16
Ignore settings for plugins that aren't supported


gilbarbara:
Hello,

Since [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) can't be used with CodeClimate, is there a way to ignore its settings? This engine breaks with `Error: unable to load resolver "webpack".` with the configuration below.

```
"settings": {
    "import/resolver": {
      "webpack": {
        "config": "./config/webpack.config.js"
      }
    }
  }
```

gusano:
any update on that?

gilbarbara:
Hey guys, any news here?
I'm unable to use CodeClimate since all my builds are broken because of this.
Is it fixable? @abbyarmada @katwchen

abbyarmada:
Hey @gilbarbara, I just pinged someone about this and we're looking into it. Thanks for your patience!

GordonDiggs:
Hey @gilbarbara I just shipped something that I think will address this! I tested it on a few of your OSS repos. Can you try another build and let me know what your experience is like?

gilbarbara:
hey @GordonDiggs 
I still got the error on codeclimate site. Is it in production yet or just the cli?
```
Ignoring the following rules that rely on module resolution:

* import/extensions
* import/no-unresolved
* import/no-extraneous-dependencies
* import/no-absolute-path
* import/no-duplicates
* import/no-named-as-default-member
module.js:471
    throw err;
    ^

Error: Cannot find module 'webpack'
...
```

GordonDiggs:
@gilbarbara can you email me a link to where you see that? gordon[at]codeclimate[dot]com

gilbarbara:
@GordonDiggs Did you get my email? Anything else I could do?

gilbarbara:
@GordonDiggs 
Just to let you know the analysis worked in the CLI (0.6.8)

gilbarbara:
@abbyarmada 
I'm getting pressured to cancel my organization subscription as we aren't getting proper reports since July. 😞

GordonDiggs:
@gilbarbara I didn't realize you are on the eslint-4 channel. Let me port the changes there!

gilbarbara:
@GordonDiggs Are these changes already on codeclimate.com runner?

GordonDiggs:
@gilbarbara they are! If you're still having trouble, I would suggest removing the parts of your config that use the resolver stuff, as it is not supported on Code Climate right now

[...]

PR #349, opened 2017-11-06
 Add more packages and blocklist more rules

GordonDiggs:
Ports #342 to the eslint-4 channel

Addresses #320


GordonDiggs:
Revisiting with the right base

Issue #370, opened 2018-01-26
Add eslint-plugin-es5 to supported plugins


smmccrohan:
Attempting to use the `eslint-plugin-es5` plugin (https://www.npmjs.com/package/eslint-plugin-es5), we're getting:

```
Module not supported: eslint-plugin-es5
/usr/local/node_modules/eslint/lib/config/config-file.js:434
            throw e;
            ^

Error: Failed to load config "plugin:es5/no-es2015" to extend from.
Referenced from: /code/.eslintrc.yml
    at configMissingError (/usr/local/node_modules/eslint/lib/config/config-file.js:193:19)
    at loadConfigFile (/usr/local/node_modules/eslint/lib/config/config-file.js:219:27)
    at load (/usr/local/node_modules/eslint/lib/config/config-file.js:535:18)
    at configExtends.reduceRight.e (/usr/local/node_modules/eslint/lib/config/config-file.js:424:36)
    at Array.reduceRight (native)
    at applyExtends (/usr/local/node_modules/eslint/lib/config/config-file.js:408:28)
    at Object.load (/usr/local/node_modules/eslint/lib/config/config-file.js:566:22)
    at loadConfig (/usr/local/node_modules/eslint/lib/config.js:63:33)
    at new Config (/usr/local/node_modules/eslint/lib/config.js:229:42)
    at CLIEngine.getConfigForFile (/usr/local/node_modules/eslint/lib/cli-engine.js:774:30)
```

Could we have the module added to the whitelist? And relatedly, is there a way to have CC recognize `extend` entries from modules that it has rejected and skip them, rather than failing?

lasseborly:
Thanks for a great product @ale7714.

This would be a really nice addition to the whitelist and would greatly benefit our multiple ES5 code bases we need to have checked across our org. Thanks again.

PR #371, opened 2018-01-30
Add eslint-plugin-es5

getabetterpic:
Adds the es5 plugin to address #370. Needed so linter can be configured to error on ES2015 features accidentally getting into ES5 files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "eslint-plugin-ejs": "^0.0.2",
     "eslint-plugin-ember": "^4.5.0",
     "eslint-plugin-ember-suave": "^1.0.0",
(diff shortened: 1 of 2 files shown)


lasseborly:
Thanks for a great product @chrishulton and everyone else at CodeClimate.

This would be a really nice addition to the whitelist and would greatly benefit our multiple ES5 code bases we need to have checked across our org. Thanks again.

chrishulton:
Thanks for the nudge.  This plugin should now be available in the stable (eslint-3) channel.  

Note, for changes to the eslint-4 channel, a PR can be made against the [channel/eslint-4 branch](https://github.com/codeclimate/codeclimate-eslint/tree/channel/eslint-4).

Issue #369, opened 2018-01-25
Add eslint-plugin-ramda to supported plugins


Undistraction:
It appears that [eslint-plugin-ramda](https://github.com/ramda/) isn't currently supported and reports errors for the rules it provides. Please add it to your whitelist.

Failures 
```
Definition for rule 'ramda/always-simplification' was not found 
Definition for rule 'ramda/compose-simplification' was not found 
etc.
```

Config:

```
---
engines:
  duplication:
    enabled: true
    config:
      languages:
      - javascript
  eslint:
    enabled: true
    checks:
          import/extensions:
            enabled: false
  fixme:
    enabled: true
ratings:
  paths:
  - "**.js"
exclude_paths:
- .idea/
- docs/
- dist/
- node_modules/
- lib/
- es/
- test/
- webpack.config.js
- webpack.config-test.web.js
- jsdoc.js
```

davehenton:
Hey @Undistraction , we just added support for `eslint-plugin-ramda` [here](https://github.com/codeclimate/codeclimate-eslint/pull/374/files) . Let me know if you run into any issues with that, or if there's anything else we can do for you.

Undistraction:
Thanks @davehenton. 

Unfortunately that [still doesn't seem to be working](https://github.com/char0n/ramda-adjunct/pull/335) for us.

davehenton:
Hey @Undistraction -- I should have noted this earlier, but we only added `eslint-plugin-ramda` to our ESLint-4 channel. To instruct Code Climate to use ESLint-4, you'll need to specifcy that channel in your `.codeclimate.yml` as directed [here](https://docs.codeclimate.com/docs/eslint#section-eslint-versions). Sorry about the confusion with that. Let me know if this gets you up and running.

Undistraction:
@davehenton No worries. And thanks.

PR #374, opened 2018-02-01
Adding support for eslint-plugin-ramda

noahd1:
Support for eslint-ramda-plugin (#369)

This has not yet been tested locally. Will update PR with status.


noahd1:
We ran a test locally:

First verifying that standard channel-2 complains if you specify a ramda rule without having the plugin, we ran against a development version of the plugin specificying a ramda rule, and the analysis didn't complain. Is there any other testing required?

noahd1:
One thing we weren't certain about was what channels we should add this plugin to ... as you can see we've only added it to the `eslint-4` channel.

Issue #390, opened 2018-05-24
Add support for CoffeeScript plugin


n-rodriguez:
Hi!
Is it possible to add support of CoffeeScript plugin? (https://www.npmjs.com/package/eslint-plugin-coffeescript)
Thank you!

larkinscott:
Hey @n-rodriguez! Adding and testing this plugin now.

n-rodriguez:
Great :)

larkinscott:
`eslint-plugin-coffeescript` has been added to `channel-3`. Closing this out!

n-rodriguez:
Awesome! Thank you!

PR #395, opened 2018-06-01
Add eslint-plugin-coffeescript

larkinscott:
Re: https://github.com/codeclimate/codeclimate-eslint/issues/390


n-rodriguez:
Hi! Do you have an idea of when it will be available in channels?

PR #402, opened 2018-07-26
Add import/no-cycle to blocklisted rules

gilbarbara:


wilson:
Hi there. Can you elaborate a little on why we need this?
Is it because the `no-cycle` check requires that the `node_modules` directory be present?

gilbarbara:
@wilson Hello! Yes, this rule requires static analysis that involves `node_modules`

wilson:
This has been merged to the ESLint 3 / master world now; this same change is probably necessary on the ESLint-4 channel as well. Which channel(s) are you using?

gilbarbara:
Cool, thanks!
I'm using `eslint-4`. Do I need to commit that change as well?

wilson:
I just pushed it to `eslint-4` as well; I'm guessing ESLint 2.x users don't encounter this newer feature, so that channel presumably doesn't need it. Thanks!

Issue #383, opened 2018-04-04
ESLint-4 prettier support.


aichholzer:
## Actual Behavior

My `.codeclimate.yml` file:
```js
version: "2"
plugins:
  eslint:
    enabled: true
    channel: "eslint-4"
exclude_patterns:
  - "**.json"
  - "**.md"
  - "**.yml"
```

when trying to run codeclimate's analysis it fails:

```js
/usr/src/app/lib/validate_config.js:18
        throw e;
        ^

TypeError: Cannot read property 'recommended' of undefined
Referenced from: /code/.eslintrc.json
    at loadConfigFile (/usr/local/node_modules/eslint/lib/config/config-file.js:211:40)
    at loadFromDisk (/usr/local/node_modules/eslint/lib/config/config-file.js:486:18)
    at load (/usr/local/node_modules/eslint/lib/config/config-file.js:550:20)
    at configExtends.reduceRight.e (/usr/local/node_modules/eslint/lib/config/config-file.js:421:36)
    at Array.reduceRight (native)
    at applyExtends (/usr/local/node_modules/eslint/lib/config/config-file.js:403:28)
    at loadFromDisk (/usr/local/node_modules/eslint/lib/config/config-file.js:514:22)
    at Object.load (/usr/local/node_modules/eslint/lib/config/config-file.js:550:20)
    at Config.getLocalConfigHierarchy (/usr/local/node_modules/eslint/lib/config.js:228:44)
    at Config.getConfigHierarchy (/usr/local/node_modules/eslint/lib/config.js:180:43)
```

[This](https://codeclimate.com/github/CometaFront/Cometa/builds/1) is the failing build.

## Expected Behavior

It should run normally and pass, since according to [this](https://github.com/codeclimate/codeclimate-eslint/blob/channel/eslint-4/package.json), you do support `eslint-config-prettier`, `eslint-plugin-prettier` and `prettier` in your ESLint-4 channel

## Steps to reproduce

This is happening on the `master` branch, therefore, please feel free to fork it. Perhaps you want to take a look at the `package.json` file.

aichholzer:
@toddmazierski -I also tried the `eslint-3` channel and [the issue](https://codeclimate.com/github/CometaFront/Cometa/builds/17) is the same.

toddmazierski:
Hi, @aichholzer! Sorry for the delay in getting back to you. I think there might be a syntax error in your `.eslintrc`. Mind trying the following change?

```patch
[...]

PR #414, opened 2018-09-20
update eslint-plugin-prettier

efueger:
update to solve typeError issue: https://github.com/codeclimate/escalations/issues/424


efueger:
@katwchen - once we get this eslint-plugin-prettier merged, I'll follow up with updating eslint-config-prettier to solve https://github.com/codeclimate/codeclimate-eslint/issues/383#issuecomment-382232888

PR #425, opened 2018-11-01
Update README.md to reference eslint-5

efueger:
@filipesperandio - is this channel ready for people to use?


filipesperandio:
@efueger Still have to enable the channel on builder and CLI. I am off (kinda) until Monday. I can get this ready by then.

Issue #430, opened 2018-11-26
eslint-5 errors out on not supported plugins


racer01:
## Actual Behavior
Errored when trying to load unsupported eslint plugin.

## Expected Behavior
Log "Module not supported" and ignore upsupported plugin.

config file: https://github.com/openHAMS/openHAMS-web2/blob/dbf0303d8e5fd03ab8bd8ae542bd16a4a8dea710/.codeclimate.yml

## Relevant links
It worked with eslint-4: https://codeclimate.com/github/openHAMS/openHAMS-web2/builds/121
but upgrading to eslint-5 broke it: https://codeclimate.com/github/openHAMS/openHAMS-web2/builds/122

filipesperandio:
This seems to be rather a change on how ESLint itself handles the scenario.
Would you be able to achieve the expected results running `eslint` itself?

racer01:
As far as I can tell [this patch](https://github.com/codeclimate/codeclimate-eslint/blob/channel/eslint-5/lib/eslint-patch.js) logs the "module not supported" message and adds the unsupported module to the skipped modules array which I suspect is broken.

filipesperandio:
@racer01 ESLint changed the way plugins are loaded. It is being a bit tricky to patch it up to support the same behavior. In the mean time, I've added the `eslint-comments` plugin to the engine. It is not yet released to the platform, but you should be able to try it out from the CLI.

racer01:
Thank you for adding `eslint-comments` plugin. I closed this issue as adding said plugin fixed my problem.

PR #431, opened 2018-11-28
Add eslint-comments plugin

filipesperandio:
Workaround for https://github.com/codeclimate/codeclimate-eslint/issues/430

There were some changes in ESLint-5 way to load plugins that is being tricky to
patch to make the engine work with non-preexisting plugins.

This adds to the engine the plugin needed by the in the issue above.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,12 @@ test: yarn.lock
 		--workdir /usr/src/app \
 		$(IMAGE_NAME) npm run $(NPM_TEST_TARGET)
 
+yarn.add:
+	docker run -ti --rm \
+		-v $(PWD):/usr/src/app \
+		--workdir /usr/src/app \
+		$(IMAGE_NAME) yarn add $(ARGS)
+
 citest:
 	docker run --rm \
 		--workdir /usr/src/app \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -26,6 +26,12 @@ test: yarn.lock
 		--workdir /usr/src/app \
 		$(IMAGE_NAME) npm run $(NPM_TEST_TARGET)
 
+yarn.add:
+	docker run -ti --rm \
+		-v $(PWD):/usr/src/app \
```
maxjacobson: We have a shell style guide somewhere that encourages us to write the long form of flags when possible to enhance readability, (tty, interactive, volume, I think, in this case).

maxjacobson: I think we have a yarn task which could be used to accomplish this goal. I don’t mind adding another with better ergonomics, but should we document it in the contributors file?

filipesperandio: yeah, we have one to generate the lock, but meaning we should change `package.json` manually, this would accomplish both things.

filipesperandio: > We have a shell style guide somewhere that encourages us to write the long form of flags when possible to enhance readability, (tty, interactive, volume, I think, in this case).

I will keep it like this to be consistent with the existing targets and assuming those are well known flags (`-ti --rm -v`)

Issue #427, opened 2018-11-07
please add eslint-import-resolver-babel-module@^5.0.0-beta1 to the eslint-5 channel


modosc:
there is prior art with some of the other resolvers (`webpack`,`node`). it would be great to have this one as well. happy to do a pr if that would speed things up?

ale7714:
@modosc thank you! A PR would definitely help :). We always welcome contributions.

modosc:
ping?

modosc:
no need now, #432 fixes this for us

Issue #426, opened 2018-11-06
eslint-import-resolver-webpack broken for eslint-5 channel


modosc:
# Bug Template

## Actual Behavior
every file generates this error on build:
```
Error resolving webpackConfig { Error: Cannot find module 'webpack'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.Module._load (module.js:475:25)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/code/webpack/config.dev.js:1:79)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3) code: 'MODULE_NOT_FOUND' }
```
and results in this new issue on every file:
```
Resolve error: Cannot find module 'webpack' NEW
```

## Expected Behavior
this doesn't happen on the eslint-4 channel. i've updated the various eslint-* modules but `eslint-import-resolver-webpack`'s version stayed the same
# Steps to reproduce
our `webpack/config.dev.js` is:
```
const webpack = require('webpack')
const merge = require('webpack-merge')
const Dotenv = require('dotenv-webpack')
const { buildCommon } = require('./common')

module.exports = merge.smart(buildCommon({ cssSourceMaps: false }), {
  devtool: 'eval',
  output: {
    filename: '[name].js',
  },
  mode: 'development',
  optimization: {
    nodeEnv: process.env.BABEL_ENV || process.env.NODE_ENV || 'development',
  },
  plugins: [
    new Dotenv(),
    new webpack.HotModuleReplacementPlugin(),
    new webpack.LoaderOptionsPlugin({
      debug: true,
    }),
  ],
})

```

## Relevant links
we are commercial customers - should i include a link to failed builds here?

noqcks:
I can replicate this bug on my own codeclimate account with a project, but I am unable to reproduce locally using the [codeclimate-cli](https://github.com/codeclimate/codeclimate). 

It looks like it was mentioned as an issue in the eslint-4 branch as well at one point https://github.com/codeclimate/codeclimate-eslint/issues/388. 

Are you able to reproduce locally @filipesperandio ? Is there anything in the way that the engines are run on the codeclimate platform that might cause this?

modosc:
[...]

PR #432, opened 2018-12-03
add setting blocklist

modosc:
i think this might solve #427 and #426 - since we're already disabling all of the `import/*` rules there's no reason to keep the `import/resolver` settings - this is what's causing these other modules to be instantiated. 

i added local specs but i don't know how to run integration specs inside of the engine/image so i haven't tested that yet.

### DIFF
--- a/lib/eslint-patch.js
+++ b/lib/eslint-patch.js
@@ -7,6 +7,7 @@ const ConfigFile = require("eslint/lib/config/config-file");
 
 const Config = require("eslint/lib/config");
 const RuleBlocklist = require("./rule_blocklist");
+const SettingBlocklist = require("./setting_blocklist");
 
 module.exports = function patch() {
   const skippedModules = [];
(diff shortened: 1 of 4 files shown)


modosc:
ping?

modosc:
ping?

filipesperandio:
@modosc I will make room for this on my schedule. Likely I will be able to work on in by Monday.
Sorry this is taking so long.

filipesperandio:
Related to #427 and #426

modosc:
thanks! this works, we can now run eslint5.

filipesperandio:
Changes are not yet in the platform, but CLI should already have it.
I am pushing this through today.

modosc:
oh - it definitely worked this morning for our builds. thanks!

filipesperandio:
It might be related to some other changes we did last week, I guess.

modosc:
hah, i'll try again just to be sure

filipesperandio:
The changes are now live in the platform :D

modosc:
oh lol, forgot to change our `.codeclimate.yml` file. changed it and all works as expected, thanks again!

PR #435, opened 2018-12-24
Update eslint-plugin-prettier and eslint-config-prettier

BPScott:
Bumps eslint-plugin-prettier from 2.6.1 to 2.7.0 to add support for honoring .prettierignore files when using eslint-plugin-prettier ([CHANGELOG](https://github.com/prettier/eslint-plugin-prettier/blob/master/CHANGELOG.md#v270-2018-09-26)). There is a v3.0.0 out but it is mostly code tidying and dropping support for unsupported versions of node (v4, v7 and v9) and ESLint (v3 and v4) which I don't think we want.

Bumps eslint-config-prettier from 3.1.0 to 3.3.0 to add support for ignoring rules supplied by eslint-plugin-vue ([CHANGELOG](https://github.com/prettier/eslint-config-prettier/blob/master/CHANGELOG.md#version-330-2018-11-11))

The latter will allow me to use state:
```
extends:[
  'plugin:vue/recommended'
  'prettier'
  'prettier/vue'
]
```

without the eslint engine complaining that it can't import from 'prettier/vue'


BPScott:
I'm using v5.

filipesperandio:
I think this makes more sense against `channel/eslint-5` channel, do you mind porting this PR over?
Does it makes sense for you?

BPScott:
Closing this, and shall reopen a new PR against channel/eslint-5.

PR #450, opened 2019-05-02
feat: add badi eslint to whitelist

jepser:
- Badi eslint configuration

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "url": "http://github.com/codeclimate/codeclimate-eslint.git"
   },
   "dependencies": {
+    "@badi/eslint-config": "^1.1.3",
     "@br/eslint-plugin-laws-of-the-game": "^1.3.0",
     "babel-eslint": "^7.2.0",
     "eslint": "^3.19.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,18 @@
 # yarn lockfile v1
 
 
+"@badi/eslint-config@^1.1.3":
+  version "1.1.3"
+  resolved "https://registry.yarnpkg.com/@badi/eslint-config/-/eslint-config-1.1.3.tgz#6f154d50ca018cf63c8adf6b20104b68e6e531ca"
+  integrity sha512-FT3IdVUJcVsaIyRvnDstsp8DKzQbsgwPzrSdSV9BcINXbJq9er7ddUYMzs2ZyKbMDKqM3Buk7ON2N5iOr19d6Q==
+  dependencies:
+    eslint-config-airbnb "^17.1.0"
+    eslint-plugin-filenames "^1.3.2"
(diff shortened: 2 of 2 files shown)


jepser:
How long does it take to get the PR reviewed @ale7714 ?

filipesperandio:
@jepser Can you elaborate on what Badi eslint brings on?

@ale7714 I am leaning towards not merging this in. The npm package in questions has only  36 downloads in total, the github link there does not exists, besides its being first released only 20 days ago.
Without any docs or even the repo, the only way to know what this does would be inspecting the source.

jepser:
@filipesperandio thanks for answering, to give you some context:
@badi we use start using a custom eslint configuration some time ago, but since the projects have been more we decided to have a  portable configuration, @badi/eslint-config.

The team installed Codeclimate on our project and we wanted to continue using it because of the value it has to us, but since we changed to a portable (private) configuration, the eslint fails since you run linting against your repository.

The eslint-config wasn't be meant to be public but since Codeclimate doesn't have a way  to install it, from a ticket we were told to make a PR with our configuration so we needed to make it public.

So, answering your question ** Can you elaborate on what Badi eslint brings on?**
It's not about 36 downloads in total, or that it's new, but the way we were told to go in order to continue using your service with our configuration.

I do agree that it needs more docs but since it's only one file with a set of rules, the time it takes to review is trivial.

If there is another option, I'm down to collaborate.

badi:
uhh, for the record this is not me. Did you mean `@badiapp`?

jepser:
Yes, sorry @badi, haha I didn't check that I would tag you. @badiapp yes.

filipesperandio:
@jepser We try to support the most common modules from the community and be flexible with the engine configuration.
Each project may have their set of configs and we don't enforce a set rules broadly.

Bringing on the @badiapp makes it broadly available, which we don't want for configuration rules, and since we can't easly inspect a private lib,
it is also risky.

I've made an example on how to configure CodeClimate to use a common eslint set of rules [here](https://github.com/filipesperandio/eslint-base-example)
that you can follow.

jepser:
Thanks @filipesperandio. I understand your point, but I don't agree on "we don't want for configuration rules", I believe a tool that provides a solution based on a broad horizontal shouldn't be opinionated. Nor "we can't easily inspect a private lib", [@badi/eslint-config](https://www.npmjs.com/package/@badi/eslint-config) is a public project and since the code is not being transpired, it should be trivial to check the file that contains the rules.

Thanks for letting us know your decision, from here we will evaluate the viability of it and decide what are we going to do.

filipesperandio:
@jepser It is not opinionated, it will be somehow if the module is incorporated, since your configs will be available to all clients, not only you.
We allow for any eslint configs as pointed out in the previous comment, this is just not the way we handle this kind of configs. Sorry if there was a confusion on our side to properly communicate that to start with.

Issue #412, opened 2018-09-06
Cascading with ESLint config files doesn't work


efueger:
# Bug Template

## Actual Behavior

Repo uses 2 .eslintrc.js files. The deeper-nested config file doesn't seem to be used with CLI or hosted results.

- Customer's using a private repo and provided this generic version for troubleshooting: https://ufile.io/12fvr\ 👌 

CLI Debug output:
```
$ CODECLIMATE_DEBUG=1 codeclimate analyze -e eslint
Starting analysis
I, [2018-09-06T02:06:12.980096 #1]  INFO -- : starting engine eslint
D, [2018-09-06T02:06:13.061906 #1] DEBUG -- : /config.json content: {"enabled"=>true, "channel"=>"eslint-4", "config"=>{"config"=>"projects/frontend/.eslintrc.js", "ignore_path"=>"projects/frontend/.eslintignore"}, "include_paths"=>["projects/", "README.md", ".codeclimate.yml"], "debug"=>"1"}
D, [2018-09-06T02:06:13.070547 #1] DEBUG -- : docker run: ["docker", "run", "--name", "cc-engines-eslint-eslint-4-72e688d0-b4f5-4d8e-8092-0b9b5d3a49d5", "--cap-drop", "all", "--label", "com.codeclimate.label=e7f857b9-175a-415a-9a21-103f9785f005", "--log-driver", "none", "--memory-swap", "-1", "--net", "none", "--rm", "--volume", "/Users/emily/workspace/code-climate-eslint:/code:ro", "--volume", "/tmp/cc/4b9bcc94-2d0f-4920-9c66-0a70862e10c9:/config.json:ro", "--user", "9000:9000", "--memory", "1024000000", "codeclimate/codeclimate-eslint:eslint-4"]
D, [2018-09-06T02:06:14.294718 #1] DEBUG -- : engine stderr: ESLint is running with the default parser.
D, [2018-09-06T02:06:14.323883 #1] DEBUG -- : engine stderr: Ignoring the following rules that rely on module resolution:
D, [2018-09-06T02:06:14.323959 #1] DEBUG -- : engine stderr: 
D, [2018-09-06T02:06:14.324959 #1] DEBUG -- : engine stderr: * import/extensions
D, [2018-09-06T02:06:14.325121 #1] DEBUG -- : engine stderr: * import/no-unresolved
[...]

PR #513, opened 2021-02-19
Add support to monorepos

fede-moya:
Some customers have being requesting support for monorepos that have more than one eslint configuration, repos that use 2 .eslintrc.js files or more. ESLint provides this feature out of the box by making use of eslint [Cascading and Hierarchy
feature](https://eslint.org/docs/user-guide/configuring/configuration-files#cascading-and-hierarchy).  

A customer shared some very interesting findings:

```
If we only have /a/.eslintrc and /b/.eslintrc (such that directory /a/ has different linting rules than directory /b/) then Code Climate throws this error when attempting to analyze our project:

Error: No ESLint configuration found in /code.at CascadingConfigArrayFactory._finalizeConfigArray 

```

And he also pointed that if we add a `.eslintrc` file containing only `{}` then the analysis goes correctly. Doing some research I found that eslint requires a config file at root level, if not given the previously mentioned error is raise. So, seems like a correct approach to use the `baseConfig` configuration of eslint's api and set it to `{}` in order to allow monorepo's analysis and do not alter existen configurations. 

Also added some end to end tests that cover this use case. 

Resolves https://github.com/codeclimate/codeclimate-eslint/issues/412


noelia-lencina (approved):
I really like this idea! Super simple and effective! 
Do you think it would be worth adding a warning message of some sort if for the cases we end up just using the default rules because no `.eslintrc` file was found anywhere?

fede-moya:
> I really like this idea! Super simple and effective!
> Do you think it would be worth adding a warning message of some sort if for the cases we end up just using the default rules because no `.eslintrc` file was found anywhere?

@noelia-lencina I don't think so. For two reasons:

PR #514, opened 2021-02-23
Update eslint-config-will-robinson

jacksonrayhamilton:
Minor update to this config that we’d like to be able to use when we lint our files.


jacksonrayhamilton:
Thank you!

PR #516, opened 2021-02-23
Add support for monorepos

fede-moya:
Same as https://github.com/codeclimate/codeclimate-eslint/pull/513 , but for eslint-5.


system-tests/mono-repo/snapshot
```diff
@@ -0,0 +1 @@
+0
```
noelia-lencina, line 1: I found it weird not to have any of this test that assert on having at least an issue. Would it be possible to raise an issue?

fede-moya: Yeah I know I tried to induce an issue was harder that what I thought. `eslint-5` seems way more permissive I don't know.

PR #525, opened 2021-04-06
Add react-perf plugin support

MosheZemah:
Add react perf support


fede-moya:
@MosheZemah the update is in place. Let me know if it worked as expected.

PR #531, opened 2021-04-27
QUA-218: Use streams to stringify large jsons

noelia-lencina:
https://codeclimate.atlassian.net/browse/QUA-218

This intends to process issues as a stream, to avoid passing a large json to `JSON.stringify`, which may cause memory problems.


lib/eslint.js
```diff
@@ -191,8 +196,12 @@ function run(console, runOptions) {
               return
             }
 
-            var issueJson = buildIssueJson(message, path)
-            STDOUT(issueJson + "\u0000\n")
+            var readableJsonStream = buildIssueJson(message, path)
+            var output = ""
```
fede-moya, line 200: Shouldn't instead be like follows ?
```ruby
            var output = "" 
            readableJsonStream.on('data', (chunk) => {
              output = output + chunk
            })
            readableJsonStream.on('end', () => process.stdout.write(`${JSON.stringify(output)}` + "\u0000\n"))
```

noelia-lencina: We want to avoid applying stringify on large jsons, that's why it's applied on the chunks, building the string little by little. Does it make sense?

fede-moya: Yeap, it does.

fede-moya (approved):

efueger:
@noelia-lencina - ready to merge?

efueger:
@noelia-lencina to fix tests

fede-moya:
**UPDATE**

I followed Noe's advice and I converted the failing integration tests into system tests. The integration tests were failing because they used to depend too much on implementation details.

PR #537, opened 2021-05-27
Add @euberdeveloper/eslint-plugin

euberdeveloper:


mileslane (changes requested):
We should not use dependencies that have a small user and developer base.  It is extremely risky, as there are many bad actors who are using such packages to install malware.  I see no reason to merge a package whose primary function, at the best, would be to simply add a bundle of other dependencies.

euberdeveloper:
Yes but this way I can not use your service with my code. I guess that this is a problem of other people, too

mileslane:
Yes, that's true.  I would be more concerned for your risk exposure. I opened a similar issue for several Babel plugins. I personally avoid using NPM packages with few users and developers. I don't want any obscure, poorly supported plugins in my code. Adding the plugin to the codeclimate-eslint package.json does not increase anyone's risk.

euberdeveloper:
In this case (to me) it's not obscure since it is written by me...

I find it as a very comfortable way to add my customized eslint configuration to all my projects and I prefer renouncing using code climate than renouncing to those configurations

euberdeveloper:
In any case small user base means only more potential damage, not that it is a bigger risk... Look what happened with colour.js and faker.js

PR #538, opened 2021-06-28
fix: adds support for @babel/preset-env

rachelslurs:


filipesperandio:
@fede-moya can you take this for a spin if you have some time?

efueger:
Hi Rachel,

Added this to the queue! Should have a reply soon.

Thanks,
Emily

PR #539, opened 2021-06-28
fix: adds support for eslint-config-next

rachelslurs:


efueger:
Hi Rachel,

Added this to the queue! Should have a reply soon.

Thanks,
Emily

PR #541, opened 2021-07-23
Upgrade: eslint to version 7.8.1

frnico33:
# Description

Upgrade eslint channel 7 to version 7.8.1

# Related issue

https://codeclimate.atlassian.net/jira/software/projects/QUA/boards/44?selectedIssue=QUA-184


fede-moya (approved):
Cool 🚀, 

still we are quite behind considering latest version is `7.31.0` but since this versions includes the desired `es2021` key I think we are good to go with this version and continue with the upgrade later.

PR #553, opened 2021-10-06
QUA-388: add next.js and update eslint-config-next

dantevvp:
To fix the error detailed in [QUA-388](https://codeclimate.atlassian.net/browse/QUA-388), the `next` package has been installed, and the `eslint-config-next` package has been updated to the matching version `11.1.2`


fede-moya (approved):
LGTM 👍🏼 , however I remember the customer reporting this issue not having a `.codeclimate.yml` file on his repo. We need to confirm which is the selected channel when enabling plugins from the UI. I assume it point's to `stable`, if that's the case then merging this change won't solve customer's problem.

PR #558, opened 2022-01-10
QUA-461: update base image

dantevvp:
This PR updates the base image from `node:6.17-stretch` to `node:17.3-bullseye-slim`.

The line `RUN apt-get update && apt-get install -y gnupg` has been added due to an error being thrown in this line:

```
RUN apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg...
```

```
gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:6.17.0-stretch
+FROM node:17.3-bullseye-slim
 LABEL maintainer "Code Climate <hello@codeclimate.com>"
 
 RUN adduser --uid 9000 --gecos "" --disabled-password app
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:6.17.0-stretch
+FROM node:17.3-bullseye-slim
```
larkinscott, line 1: :raised_hands: were you able to build the image locally and test it via the CLI?

dantevvp: yes! CLI returned no errors and a lot less vulnerabilities

larkinscott: Awesome

Issue #567, opened 2022-02-02
Add support for plugins for eslint-7 channel


mileslane:
## Actual Behavior

It would be helpful to be able to get our GitHub Codeclimate checks to pass.  Currently, we are getting errors like this:

`Parsing error: Cannot find module 'babel-plugin-macros' Require stack: - /code/babel.config.js`

## Expected Behavior

Our babel.config.js file uses the following plugins that are not present in the list of Dependencies in the eslint-7 package.json file:

- babel-plugin-macros
- @babel/preset-typescript
- @babel/plugin-proposal-optional-chaining
- @babel/plugin-syntax-dynamic-import
- @babel/plugin-transform-destructuring
- @babel/plugin-proposal-class-properties
- @babel/plugin-proposal-private-methods
- @babel/plugin-proposal-private-property-in-object
- @babel/plugin-proposal-object-rest-spread
- @babel/plugin-transform-regenerator

## Relevant links

----
- [X] 4-6 word issue title

mileslane:
Here is my proposed patch (I don't have write access):
diff --git a/package.json b/package.json
index 39296eac..2e871ec3 100644
--- a/package.json
+++ b/package.json
@@ -15,11 +15,20 @@
     "@babel/plugin-transform-runtime": "^7.13.10",
     "@babel/preset-env": "^7.14.7",
     "@babel/preset-react": "^7.14.5",
+    "@babel/preset-typescript": "^7.16.7",
+    "@babel/plugin-proposal-object-rest-spread": "^7.16.7",
+    "@babel/plugin-proposal-optional-chaining": "^7.16.7",
+    "@babel/plugin-proposal-private-methods": "^7.16.11",
+    "@babel/plugin-proposal-private-property-in-object": "^7.16.7",
+    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
+    "@babel/plugin-transform-destructuring": "^7.16.7",
+    "@babel/plugin-transform-regenerator": "^7.16.7",
     "@babel/runtime": "^7.14.8",
     "@hapi/eslint-plugin": "^5.0.0",
     "@typescript-eslint/eslint-plugin": "^4.33.0",
     "@typescript-eslint/parser": "^4.33.0",
     "babel-eslint": "^10.0.1",
+    "babel-plugin-macros": "^3.1.0",
     "babel-plugin-react-generate-property": "^1.0.5",
     "eslint": "7.8.1",
[...]

PR #578, opened 2022-06-13
QUA-517: add multiple babel plugins (eslint-8)

dantevvp:
This PR installs the following babel plugins to the eslint-8 channel:

- babel-plugin-macros
- @babel/preset-typescript
- @babel/plugin-proposal-optional-chaining
- @babel/plugin-syntax-dynamic-import
- @babel/plugin-transform-destructuring
- @babel/plugin-proposal-class-properties
- @babel/plugin-proposal-private-methods
- @babel/plugin-proposal-private-property-in-object
- @babel/plugin-proposal-object-rest-spread
- @babel/plugin-transform-regenerator

https://codeclimate.atlassian.net/browse/QUA-517


camillof:
Lot of questions over here, this is a new world for me:
- Not sure which criteria do we use to bump the [version here](https://github.com/codeclimate/codeclimate-eslint/pull/578/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R3) but may be a good opportunity to do it? 🤔 
- In the description the `babel-plugin-macros` is listed as added, is it actually?

And I would also like to bring up this comment in the original issue: https://github.com/codeclimate/codeclimate-eslint/issues/567#issuecomment-1030643030

> Ah. I see the problem. Our lead developer is requiring individual plugins from @babel/preset-env in the process of specifying the babel configuration.
This appears to be the issues he is trying to resolve:
Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-property-in-object since the "loose" mode option was set to "true" for @babel/plugin-proposal-private-methods.
Is it possible to make Codeclimate not choke when individual plugins are required that are included in @babel/preset-env?

Is this issue solved just by adding the `plugin-proposal-private-methods` ? According to the [official docs ](https://babeljs.io/docs/en/babel-plugin-proposal-private-methods) it's already included on `@babel/preset-env` which we already have. We don't need to specify the `loose` [option](https://babeljs.io/docs/en/babel-plugin-proposal-private-methods#loose)? 

Sorry if there are too many questions!

dantevvp:
> Lot of questions over here, this is a new world for me:
> ...

@camillof I added `babel-plugin-macros` and it looks like another plugin was suggested in place for the other ones I added here, so I'll analyze adding `babel-plugin-transform-react-remove-prop-types`.

Regarding the "loose" warning, it looks like it's solved by adding only `plugin-proposal-private-methods` like you're saying, so I'll be trying that out

